PR #4, opened 2024-01-29
add an experimental rust query parser

karakanb:


karakanb:
we have decided to go with sqlglot, closing this one.

PR #6, opened 2024-02-05
Patch/add redshift

albertobruin:


pkg/config/manager.go
```diff
@@ -14,6 +14,17 @@ import (
...
 
+type PostgresConnection struct {
+	Name         string `yaml:"name" json:"name"`
+	Username     string `yaml:"username" json:"username"`
+	Password     string `yaml:"password" json:"password"`
+	Host         string `yaml:"host" json:"host"`
+	Port         int    `yaml:"port" json:"port"`
+	Database     string `yaml:"database" json:"database"`
+	PoolMaxConns int    `yaml:"pool_max_conns" json:"pool_max_conns" default:"10"`
+	SslMode      string `yaml:"ssl_mode" json:"ssl_mode" default:"disable"`
```
karakanb, line 25: do we need json counterparts for these? I think it's just YAML

pkg/postgres/db_test.go
```diff
@@ -0,0 +1,118 @@
...
+	tests := []struct {
+		name      string
+		query     string
+		expected  string
+		setupMock func(mock pgxmock.PgxPoolIface)
+		wantErr   string
+		want      [][]interface{}
+	}{
+		{
+			name:    "Test Select rows",
```
karakanb, line 27: nitpick, could you please make these all lowercase to be consistent with all the other tests?

karakanb (approved):

pkg/config/manager.go
albertobruin: Well I'm not sure. The other config types don't have json annotations but they have a `MarshalJSON` [method](https://github.com/bruin-data/bruin/blob/ff74b51e052cf6d0bbd9589af07a61316c9c7be7/pkg/config/manager.go#L72) so I thought it was  necessary?

pkg/postgres/db_test.go
albertobruin: sure thing

PR #11, opened 2024-02-14
Add MS SQL Support

albertobruin:


cmd/run.go
```diff
@@ -408,6 +409,24 @@ func setupExecutors(s *scheduler.Scheduler, config *config.Config, conn *connect
...
+		msCheckRunner := mssql.NewColumnCheckOperator(conn)
+
+		msCustomCheckRunner := ansisql.NewCustomCheckOperator(conn)
+
+		mainExecutors[pipeline.AssetTypeMsSQLQuery][scheduler.TaskInstanceTypeMain] = msOperator
+		mainExecutors[pipeline.AssetTypeMsSQLQuery][scheduler.TaskInstanceTypeColumnCheck] = msCheckRunner
+		mainExecutors[pipeline.AssetTypeMsSQLQuery][scheduler.TaskInstanceTypeCustomCheck] = msCustomCheckRunner
+
+		// we set the Python runners to run the checks on MsSQL
+		if estimateCustomCheckType == pipeline.AssetTypeMsSQLQuery {
```
karakanb, line 424: don't forget to add this to the possible alternatives to the estimator a few lines above

pkg/config/manager.go
```diff
@@ -26,6 +26,16 @@ type PostgresConnection struct {
 	SslMode      string `yaml:"ssl_mode" json:"ssl_mode" default:"disable"`
 }
 
+type MsSQLConnection struct {
+	Name     string `yaml:"name" json:"name"`
+	Username string `yaml:"username" json:"username"`
+	Password string `yaml:"password" json:"password"`
+	Host     string `yaml:"host"     json:"host"`
+	Port     int    `yaml:"port"     json:"port"`
+	Instance string `yaml:"instance" json:"instance"`
```
karakanb, line 35: what's an instance? is there a way we can eliminate this and make it similar to other db connections? is that schema maybe? thinking outloud.

pkg/mssql/materialization.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func buildCreateReplaceQuery(task *pipeline.Asset, query string) (string, error) {
+	mat := task.Materialization
+
+	if len(mat.ClusterBy) > 0 {
+		return "", errors.New("MsSQL assets do not support `cluster_by`")
+	}
+
+	return fmt.Sprintf("DROP TABLE IF EXISTS %s; SELECT tmp.* INTO %s FROM (%s) AS tmp;", task.Name, task.Name, query), nil
```
karakanb, line 48: this should prolly be wrapped in a transaction

pkg/config/manager.go
albertobruin: Instance is an [MSSQL concept](task.Name, query) and equates to `database` in other RDBMS. Its unfortunate because there is also `database` and even `schema` it seems, although I don't see a way to specify the latter. Do you think we can handle them differently?

PR #13, opened 2024-02-19
add support for synapse

albertobruin:


pkg/pipeline/pipeline.go
```diff
@@ -24,6 +24,7 @@ const (
 	AssetTypePostgresQuery        = AssetType("pg.sql")
 	AssetTypeRedshiftQuery        = AssetType("rs.sql")
 	AssetTypeMsSQLQuery           = AssetType("ms.sql")
+	AssetTypeSynapseQuery         = AssetType("sy.sql")
```
karakanb, line 27: I don't think `sy.sql` is obvious for synapse, shall we go with `synapse.sql`, @sabrikaragonen what are your thoughts?

sabrikaragonen: i agree, sy is very weird

pkg/synapse/materialization_test.go
```diff
@@ -0,0 +1,188 @@
...
+			task: &pipeline.Asset{
+				Name: "my.asset",
+				Materialization: pipeline.Materialization{
+					Type: pipeline.MaterializationTypeView,
+				},
+			},
+			query: "SELECT 1",
+			want: []string{
+				"^DROP VIEW IF EXISTS my\\.asset;$",
+				"^CREATE VIEW my\\.asset AS SELECT 1;$",
```
sabrikaragonen, line 37: here a single statement would solve it I guess, CREATE OR ALTER VIEW

pkg/synapse/materialization_test.go
```diff
@@ -0,0 +1,188 @@
...
+			task: &pipeline.Asset{
+				Name: "my.asset",
+				Materialization: pipeline.Materialization{
+					Type: pipeline.MaterializationTypeTable,
+				},
+			},
+			query: "SELECT 1",
+			want: []string{"^SELECT tmp\\.\\* INTO #bruin_tmp_.+ FROM \\(SELECT 1\\) AS tmp;\n" +
+				"IF OBJECT_ID\\('my\\.asset', 'U'\\) IS NOT NULL DROP TABLE my\\.asset;\n" +
+				"SELECT \\* INTO my\\.asset FROM #bruin_tmp_.+;$"},
```
sabrikaragonen, line 51: deleting temp table after select into might be useful

pkg/synapse/materialization_test.go
albertobruin: There is no `CREATE OR ALTER VIEW...` in synapse sql, despite it existing in ms sql

pkg/synapse/materialization_test.go
albertobruin: Missed that. Thanks

PR #17, opened 2024-03-06
Feature/add ingestr

albertobruin:


pkg/ingestr/operator.go
```diff
@@ -0,0 +1,74 @@
...
+func (o BasicOperator) Run(ctx context.Context, ti scheduler.TaskInstance) error {
+	return o.RunTask(ctx, ti.GetPipeline(), ti.GetAsset())
+}
+
+func (o BasicOperator) RunTask(ctx context.Context, p *pipeline.Pipeline, t *pipeline.Asset) error {
+	return errors.New("Ingestr tasks not implemented")
+}
+
+type IngestrCheckOperator struct {
+	configs *map[pipeline.AssetType]executor.Config
```
karakanb, line 27: why is the value here of type `executor.Config` but not an actual column check operator instance?

pkg/ingestr/operator.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (i IngestrCheckOperator) Run(ctx context.Context, ti scheduler.TaskInstance) error {
+	asset := ti.GetAsset()
+	assetType, err := asset.GetIngestrDestinationType()
```
karakanb, line 32: I think this is very specific to the operator itself, I wouldn't put this method on the asset itself

pkg/ingestr/operator.go
albertobruin: The idea is that the ingestr check operator needs to be able to perform checks for different types of destination. To illustrate with an example, imagine you have a pipeline where you have 2 ingestr assets, one has a postgres destination and another one has a mssql destination. The are both ingestr so they would use this check operator but the checks need to be different

pkg/ingestr/operator.go
albertobruin: Thing is that it's needed not only in the operator but also outside so I thought it was a nice abstraction. Will create a function function in the `ingestr` package instead like

```go
func GetIngestrDestinationType(asset Asset) (AssetType, error)
```

PR #29, opened 2024-04-04
simplify ingestr yaml

albertobruin:


pkg/ingestr/operator.go
```diff
@@ -58,35 +58,38 @@ func (o BasicOperator) Run(ctx context.Context, ti scheduler.TaskInstance) error
...
+
 	sourceTable, ok := ti.GetAsset().Parameters["source_table"]
 	if !ok {
 		return errors.New("source table not configured")
 	}
 
-	destConnectionName, ok := ti.GetAsset().Parameters["destination_connection"]
-	if !ok {
-		return errors.New("destination connection not configured")
+	destConnectionName := ti.GetAsset().Connection
```
karakanb, line 77: this wouldn't work, we need this to be a function that can get the pipeline-level defaults if needed as well

PR #40, opened 2024-04-23
Feature/add notion connection

albertobruin:


examples/simple-pipeline/assets/hello_ingestr5_asset.yml
```diff
@@ -0,0 +1,9 @@
+name: public.test_merge_5
+type: ingestr
+connection: neon
+
+parameters:
+  source_connection: localmysql
+  source_table: 'test.Persons'
+
+  destination: postgres
```
karakanb, line 9: why include this here?

pkg/config/manager.go
```diff
@@ -62,6 +62,11 @@ type MySQLConnection struct {
 	Database string `yaml:"database" json:"database"`
 }
 
+type NotionConnection struct {
+	Name  string `yaml:"name" json:"name"`
+	Token string `yaml:"token" json:"token"`
```
karakanb, line 67: we call this `api_key` in ingestr, maybe we can use the same lingo here?

karakanb (approved):

PR #56, opened 2024-06-19
add gorgias and shopify connection

albertobruin:


pkg/shopify/config.go
```diff
@@ -0,0 +1,14 @@
...
+	"fmt"
+)
+
+type Config struct {
+	URL    string
+	APIKey string
+}
+
+func (c *Config) GetIngestrURI() string {
+	return fmt.Sprintf("gorgias://%s?api_key=%s", c.URL, c.APIKey)
```
karakanb, line 13: this seems like a copy paste mistake?
```suggestion
	return fmt.Sprintf("shopify://%s?api_key=%s", c.URL, c.APIKey)
```

pkg/gorgias/config.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+type Config struct {
+	Domain string
+	APIKey string
+	Email  string
+}
+
+func (c *Config) GetIngestrURI() string {
+	return fmt.Sprintf("gorgias://%s?api_key=%s&%s", c.Domain, c.APIKey, c.Email)
```
karakanb, line 14: seems like we are missing the key for the email:

```suggestion
	return fmt.Sprintf("gorgias://%s?api_key=%s&email=%s", c.Domain, c.APIKey, c.Email)
```

PR #71, opened 2024-08-15
athena wip

albertobruin:


pkg/athena/db.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+func (db *DB) GetResultsLocation() string {
+	return db.config.OutputBucket
+}
+
+func (db *DB) RunQueryWithoutResult(ctx context.Context, query *query.Query) error {
+	_, err := db.Select(ctx, query)
+	return err
```
karakanb, line 33: you need to call `initializeDB` here as well just in case

pkg/athena/materializer.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/bruin-data/bruin/pkg/helpers"
+	"github.com/bruin-data/bruin/pkg/pipeline"
+)
+
+// The other packages all use a materializer that renders the query to a single string. Due to the quirks of athena
+// we need to create a different materializer that returns a slice of strings, since databricks server requires us to send separate batches
```
karakanb, line 12:
```suggestion
// we need to create a different materializer that returns a slice of strings, since athena server requires us to send separate batches
```

pkg/athena/materializer.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func NewRenderer(fullRefresh bool) *Renderer {
+	return &Renderer{
+		mat: NewMaterializer(fullRefresh),
+	}
+}
+
+func (r *Renderer) Render(asset *pipeline.Asset, query string) (string, error) {
+	queries, err := r.mat.Render(asset, query, "s3://{output bucket}")
```
karakanb, line 58: not sure I understand what you are doing here, I thought we were gonna take the connection if it exists, if not use this fake location, no?

cmd/run.go
```diff
@@ -501,6 +502,20 @@ func setupExecutors(s *scheduler.Scheduler, config *config.Config, conn *connect
 		mainExecutors[pipeline.AssetTypeIngestr][scheduler.TaskInstanceTypeCustomCheck] = ingestrCustomCheckRunner
 	}
 
+	if s.WillRunTaskOfType(pipeline.AssetTypeAthenaQuery) || estimateCustomCheckType == pipeline.AssetTypeAthenaQuery {
+		athenaOperator := athena.NewBasicOperator(conn, wholeFileExtractor, athena.NewMaterializer(fullRefresh))
```
karakanb, line 506: how come you use a regular `conn` object, since it doesn't return a DB instance for athena?

pkg/connection/connection.go
```diff
@@ -399,6 +428,24 @@ func (m *Manager) AddSfConnectionFromConfig(connection *config.SnowflakeConnecti
...
+	if m.Aws == nil {
+		m.Aws = make(map[string]*athena.DB)
+	}
+
+	m.Aws[connection.Name] = athena.NewDB(&athena.Config{
+		Region:          connection.Region,
+		OutputBucket:    connection.QueryResultsPath,
+		AccessID:        connection.AccessKey,
+		SecretAccessKey: connection.SecretKey,
+	})
```
karakanb, line 444: we don't really want to do this, that's why my previous implementation used a separate connection fetcher that created a different instance. how would we implement things like S3 sensors if we do this?

pkg/helpers/helpers.go
```diff
@@ -24,6 +25,11 @@ func GetIngestrDestinationType(asset *pipeline.Asset) (pipeline.AssetType, error
 }
 
 func PrefixGenerator() string {
+	// Always return same when testing
+	if flag.Lookup("test.v") != nil {
+		return "abcefghi"
+	}
```
karakanb, line 31: why not create another function and use that for the tests instead?

pkg/athena/materializer.go
albertobruin: I missed this. let me merge this PR and I will address this in a separate one

pkg/connection/connection.go
albertobruin: I think we can change it as soon as we have aws connections for another integration?

pkg/helpers/helpers.go
albertobruin: This was quicker and required less changes in the code. I thought it was smart

cmd/run.go
albertobruin: I imagine this comment goes together with https://github.com/bruin-data/bruin/pull/71#discussion_r1728723693

PR #87, opened 2024-09-27
Add stripe connection

sanjushahgupta:
Below changes are made in this PR:
- added  StripeConnection struct and method to return the connection's name.
- added/removed Stripe connection from a specific environment by name
- added stripe option in pipeline.go
- added config and db file to klaviyo package

### DIFF
--- a/examples/simple-pipeline/assets/hello_ingestr7_asset.yml
+++ b/examples/simple-pipeline/assets/hello_ingestr7_asset.yml
@@ -3,7 +3,7 @@ type: ingestr
 connection: postgres
 
 parameters:
-  source_connection: new_stripe
+  source_connection: my-stripe
   source_table: 'customer'
 
   destination: postgres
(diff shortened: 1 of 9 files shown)


examples/simple-pipeline/assets/hello_ingestr7_asset.yml
```diff
@@ -0,0 +1,9 @@
+name: public.stripe
+type: ingestr
+connection: postgres
+
+parameters:
+  source_connection: new_stripe
```
albertobruin, line 6: maybe for the sake of not making it confusing, let's give the source connection the same name as [here](https://github.com/bruin-data/bruin/pull/87/files#diff-d3df4c32515d23a56e039eb1897c46b80d2ddcab4eca491282fcab99b47695e6R21)

sanjushahgupta: I made changes and pushed it.

PR #92, opened 2024-10-10
Add Lint Check for Duplicate Column Names in Pipelines

tqindia:
This PR introduces a new lint check for pipelines to validate and catch duplicate column names.

- [ ] Unit Test
- [ ] Docs


pkg/lint/rules.go
```diff
@@ -343,6 +343,38 @@ func EnsurePipelineStartDateIsValid(p *pipeline.Pipeline) ([]*Issue, error) {
...
+//
+// The function performs a case-insensitive comparison of column names.
+//
+// Parameters:
+//   - p: A pointer to the pipeline.Pipeline struct containing the assets to be validated.
+//
+// Returns:
+//   - A slice of *Issue, each describing a duplicate column name found.
+//   - An error, which is always nil in this implementation.
+func ValidateDuplicateColumnNames(p *pipeline.Pipeline) ([]*Issue, error) {
```
karakanb, line 357: I'd suggest making this a single asset validator, so that we can use the same check both when the user is validating only a single asset, as well as when a full pipeline is being validated.

you can see an example implementation for another validator here: https://github.com/bruin-data/bruin/blob/main/pkg/lint/list.go#L50-L51

basically, once we have the function that does the validation only on a single asset, the `CallFuncForEveryAsset` function would make it run for every asset in the pipeline as well.

does that make sense?

PR #94, opened 2024-10-15
Implemented a cli command for testing connections

Barbar1432:
## Summary
This PR adds a new CLI command `bruin connections test` to test the validity of a connection. The command runs a `SELECT 1` query to verify if the connection is working.

## Usage
```bash
bruin connections test --environment <env> --type snowflake --name <connection-name> --output <plain|json>

### CODE REVIEW

File: cmd/connections.go [hunk 1]
```diff
@@ -276,3 +280,100 @@ func printErrorForOutput(output string, err error) {
...
+				Query := query.Query{
+					Query: "SELECT 1", // Your SQL query here
+				}
+
+				err = db.RunQueryWithoutResult(context.Background(), &Query)
+				if err != nil {
+					printErrorForOutput(output, errors2.Wrap(err, "failed to run test query on Snowflake connection"))
+					return cli.Exit("", 1)
+				}
+				fmt.Printf("Successfully tested Snowflake connection: %s in environment: %s\n", name, environment)
```
karakanb, line 356: this should go into a separate method in the snowflake database struct instead, sth like `Test()` method

Issue #96, opened 2024-10-15
'Panic: Panic' Error When JSON Keyfile Is Missing or Incorrect


DjamilaBaroudi:
## Description:
A panic: panic error occurs when trying to add a BigQuery connection without providing a valid JSON keyfile or providing an incorrect one. The error message is as follows:

1. For no file provided : 

```bash
panic: panic: failed to add BigQuery connection 'connection-name': no credentials provided
```

2. For incorrect file name
```bash
panic: panic: failed to add BigQuery connection 'connection-name': failed to create bigquery client: bigquery: constructing client: cannot read credentials file: open keyfilesz.json: no such file or directory 
```

## Steps to Reproduce:
  1. Attempt to add a BigQuery connection without providing a JSON keyfile or providing a wrong file path in the `.bruin.yml` file.
  2. Run the command `bruin validate  -o json .`
  3. Observe the panic: panic error message in the logs.

## Expected Behaviour:
- The system should gracefully handle the error by displaying a user-friendly message such as "No credentials provided" or "Invalid credentials file path," without crashing.

- If the `-o json` flag is added to the validate command, the error should be returned in JSON format instead of a plain text error message.

y-bruin:
@albertobruin @DjamilaBaroudi We have a panic statement for all types of connections. Should I address panic handling for all sources, or should I only focus specifically on BigQuery? https://github.com/y-bruin/bruin/blob/41429dc096ee1be753b1d635ec3d39c68ce93b7a/pkg/connection/connection.go#L1113

albertobruin:
so I would say, if you can catch all the panics with a catch-all then do so (im not sur ethis is possible though)
If you cannot, just handle the ones that come from adding a connection like you've done here for now cc @karakanb

PR #100, opened 2024-10-21
Hotfix: handle Panic in connection manager

y-bruin:
fix: https://github.com/bruin-data/bruin/issues/96

### DIFF
--- a/pkg/connection/connection.go
+++ b/pkg/connection/connection.go
@@ -959,29 +959,37 @@ func NewManagerFromConfig(cm *config.Config) (*Manager, error) {
 	connectionManager := &Manager{}
 
 	var wg conc.WaitGroup
+
+	var errList []error
+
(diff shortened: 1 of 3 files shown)


pkg/connection/connection.go
```diff
@@ -959,65 +959,77 @@ func NewManagerFromConfig(cm *config.Config) (*Manager, error) {
...
+
+	var errList []error
+
 	for _, conn := range cm.SelectedEnvironment.Connections.AthenaConnection {
 		wg.Go(func() {
 			err := connectionManager.AddAthenaConnectionFromConfig(&conn)
+
 			if err != nil {
-				panic(errors.Wrapf(err, "failed to add AWS connection '%s'", conn.Name))
+				errList = append(errList, errors.Wrapf(err, "failed to add AWS connection '%s'", conn.Name))
```
albertobruin, line 970: I think you need to safeguard `errList` with a mutex

PR #103, opened 2024-10-22
Migrate goreleaser from v1 to v2 

albertobruin:
- Migrate goreleaser from v1 to v2
- Skip windows ARM build from goreleaser

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -41,3 +41,22 @@ jobs:
           go-version-file: 'go.mod'
 
       - run: make format
+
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test.yml
```diff
@@ -41,3 +41,22 @@ jobs:
           go-version-file: 'go.mod'
 
       - run: make format
+
```
y-bruin, line 44: @albertobruin I added a go releaser build in PR, Is it ok ?

y-bruin (approved):

PR #123, opened 2024-10-29
Release workflow for CGO Build

y-bruin:
Test: https://github.com/y-bruin/bruin/actions/runs/11572412765
Release: https://github.com/y-bruin/bruin/releases/tag/v0.0.9


.github/workflows/build-test.yml
```diff
@@ -42,20 +42,40 @@ jobs:
...
         with:
-          install-only: true
           version: latest
+          args: build --clean --snapshot --single-target --id bruin-windows-amd64
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ github.ref_name }}
+
+
+  goreleaser:
```
karakanb, line 70: I'd call this `goreleaser-unix` to make it explicit

.github/workflows/release.yml
```diff
@@ -5,22 +5,136 @@ on:
...
           version: latest
+          distribution: goreleaser-pro
+          args: release --clean --split
+        env:
+          GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+          GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
+          VERSION: ${{ github.ref_name }}
+          GGOOS: windows
+
+  prepare-darwin:
```
karakanb, line 45: for after we merge this: could we make the `prepare-darwin` and `prepare-linux` ones to be a matrix-build instead, since the blocks seem almost the same?

.github/workflows/build-test.yml
y-bruin: done

PR #133, opened 2024-10-30
Fix add bq connection from config

Barbar1432:
Refined BigQuery connection setup by removing some of thechecks while adding minimal, effective validation for credentials.

### DIFF
--- a/pkg/bigquery/db.go
+++ b/pkg/bigquery/db.go
@@ -139,6 +139,45 @@ func (d *Client) Select(ctx context.Context, query *query.Query) ([][]interface{
 	return result, nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/bigquery/db.go
```diff
@@ -139,6 +139,45 @@ func (d *Client) Select(ctx context.Context, query *query.Query) ([][]interface{
 	return result, nil
 }
 
+func (d *Client) SelectWithSchema(ctx context.Context, query *query.Query) ([]string, [][]interface{}, error) {
```
karakanb, line 142: let's make the response to be a struct like:
```go
type QueryResult struct {
    Columns []string
    Rows []interface{}
}
```

that way the function becomes more self-explanatory, and gives us the ability to expand response structures, e.g. adding column types in the future, without having to rewrite all the function/interface signatures.

PR #157, opened 2024-11-05
 Add interactive UX in init command

y-bruin:
- If user doesn't the template name as argument 

https://github.com/user-attachments/assets/fc27a7a9-051a-4cb3-be22-27946f20a3d8

- If user pass the template name as argument 
Screenshot 2024-11-07 at 10 32 43 AM


cmd/init.go
```diff
@@ -20,6 +21,59 @@ const (
...
+				m.cursor = len(choices) - 1
+			}
+		}
+	}
+	// Return the concrete type instead of the interface
+	return m, nil // This line is fine as it is
+}
+func (m model) View() string {
+	s := strings.Builder{}
+	s.WriteString("Please select the template?\n\n")
```
karakanb, line 61:
```suggestion
	s.WriteString("Please select a template below\n\n")
```

cmd/init.go
```diff
@@ -20,6 +21,59 @@ const (
...
+
+	for i, choice := range choices {
+		if m.cursor == i {
+			s.WriteString("(•) ")
+		} else {
+			s.WriteString("( ) ")
```
karakanb, line 67: I would indent these a bit:
```suggestion
			s.WriteString(" [x] ")
		} else {
			s.WriteString(" [ ] ")
```

templates/templates.go
```diff
@@ -7,3 +7,16 @@ import (
 //go:embed *
 //go:embed */.bruin.yml
 var Templates embed.FS
+
+func TemplatesName() []string {
```
karakanb, line 11:
```suggestion
func TemplateNames() []string {
```

karakanb (approved):

PR #164, opened 2024-11-06
added validation for connection name

Barbar1432:
Connection name validation to check if it contains "-" and gives an error if yes


pkg/config/manager.go
```diff
@@ -680,3 +690,31 @@ func removeConnection[T interface{ GetName() string }](connections []T, name str
 	}
 	return connections
 }
+func (c *Connections) ValidateConnectionNames() error {
+	val := reflect.ValueOf(c).Elem() // Get the actual value that c points to
```
albertobruin, line 694: is there any way we can do this without reflections?

karakanb:
let's park this one for now.

PR #168, opened 2024-11-07
Added docs for CICD integration 

y-bruin:


.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,38 @@
+name: Build Docs
```
karakanb, line 1: we don't need this, the CI pipeline already builds and deploys docs

PR #172, opened 2024-11-07
Validate connections before running

albertobruin:
before `bruin run` validate all connections necessary are present to avoid lots of errors on big pipelines

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -259,7 +259,7 @@ func Run(isDebug *bool) *cli.Command {
 
 			err = cm.CanRunPipeline(foundPipeline)
 			if err != nil {
-				errorPrinter.Printf("Cannot run: %v\n", err)
+				errorPrinter.Printf(err.Error()) // nolint: govet
 				return cli.Exit("", 1)
 			}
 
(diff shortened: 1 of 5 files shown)


pkg/config/manager_test.go
```diff
@@ -701,3 +702,136 @@ func TestDeleteConnection(t *testing.T) {
...
+				},
+			},
+			pipeline: pipeline.Pipeline{
+				Assets: []*pipeline.Asset{
+					{
+						Connection: "conn1",
+					},
+					{
+						Connection: "conn2",
+					},
```
karakanb, line 818: I would add some test cases where the asset does not have the connection, e.g. if `bq.sql` asset is running, there likely won't be a connection, so it'll rely on the default connections of the pipeline, and this change should work with those too

cmd/run.go
```diff
@@ -257,6 +257,12 @@ func Run(isDebug *bool) *cli.Command {
 				infoPrinter.Printf("Running only the asset '%s'\n", task.Name)
 			}
 
+			err = cm.CanRunPipeline(foundPipeline)
+			if err != nil {
+				errorPrinter.Printf("Cannot run: %v\n", err)
```
karakanb, line 262: why not print the error directly? it seems nice enough, no?

PR #177, opened 2024-11-08
shopify tutorial

y-bruin:


templates/shopify-bigquery/assets/shopify.asset.yml
```diff
@@ -0,0 +1,8 @@
+name: raw.shopify
```
sabrikaragonen, line 1: It's an orders table, so please name it orders. Also, we need other source_tables as well. So for each source_table, we need an asset

sabrikaragonen: And instead of raw, maybe we can call it shopify_raw.orders

templates/shopify-bigquery/assets/shopify.asset.yml
```diff
@@ -0,0 +1,8 @@
...
+type: ingestr
+
+parameters:
+  source_connection: my-shopify-connection
+  source_table: orders
+  destination_connection: gcp
```
sabrikaragonen, line 7: If we defined it in default connections, we don't need destination connection

templates/shopify-bigquery/assets/shopify.asset.yml
y-bruin: done

PR #191, opened 2024-11-12
Add uv ingestr runner

albertobruin:


pkg/ingestr/operator.go
```diff
@@ -271,21 +154,21 @@ func (o *BasicOperator) ConvertTaskInstanceToContainerConfig(ctx context.Context
...
+
+	if strings.HasPrefix(destURI, "duckdb://") {
+		duck.LockDatabase(destURI)
+		defer duck.UnlockDatabase(destURI)
+	}
+
+	if strings.HasPrefix(sourceURI, "duckdb://") {
+		duck.LockDatabase(sourceURI)
+		defer duck.UnlockDatabase(sourceURI)
 	}
```
karakanb, line 171: this could get into a locked situation if the user is moving data within the same db?

PR #206, opened 2024-11-13
Added shopify docs and fix template

y-bruin:


docs/.vitepress/config.mjs
```diff
@@ -140,6 +140,13 @@ export default defineConfig({
                     {text: "Github Actions", link: "/cicd/github-action"},
                 ],
             },
+            {
+                text: "Connection",
+                collapsed: false,
+                items: [
+                    {text: "Shopify", link: "/connection/shopify"},
+                ],
+            },
```
karakanb, line 149: I'd add this under the sources inside Data ingestion section

PR #209, opened 2024-11-14
Add new flag for column lineage

y-bruin:
Asset1 
```
/* @bruin
name: example
type: duckdb.sql
materialization:
   type: table

depends:
  - example2

@bruin */

select * from example2;

```

Asset2
```
/* @bruin
name: example2
type: duckdb.sql
materialization:
   type: table

columns:
  - name: id
    type: integer
    description: "Just a number"
    primary_key: true
    checks:
        - name: not_null
        - name: positive
        - name: non_negative
  - name: country
    type: varchar
    description: "the country"
    primary_key: true
    checks:
        - name: not_null
  - name: name
    type: varchar
    update_on_merge: true
    description: "Just a name"
    checks:
        - name: unique
        - name: not_null
#        - name: pattern
#        value: "^[A-Z][a-z]*$"
   @bruin */

select id,name, country from example2;
```

Test
```
bruin internal parse-asset example.sql
```

Output: https://jsonblob.com/1306646592982212608

### DIFF
--- a/cmd/internal.go
+++ b/cmd/internal.go
@@ -259,13 +259,13 @@ func ParseLineage(pipe *pipeline.Pipeline, asset *pipeline.Asset) error {
 	for _, lineageCol := range lineage.Columns {
 		for _, upstream := range lineageCol.Upstream {
 			upstreamAsset := pipe.GetAssetByName(upstream.Table)
-			if upstreamAsset == nil {
-				return fmt.Errorf("upstream asset not found: %s", upstream.Table)
-			}
 
+			if upstreamAsset == nil {
+				continue
+			}
 			upstreamCol := upstreamAsset.GetColumnWithName(upstream.Column)
 			if upstreamCol == nil {
-				return fmt.Errorf("column %s not found in asset %s", upstream.Column, upstream.Table)
+				continue
 			}
 
 			newCol := *upstreamCol
--- a/cmd/internal.go
+++ b/cmd/internal.go
@@ -273,8 +273,6 @@ func ParseLineage(pipe *pipeline.Pipeline, asset *pipeline.Asset) error {
 
 			if col := asset.GetColumnWithName(lineageCol.Name); col == nil {
 				asset.Columns = append(asset.Columns, newCol)
-			} else {
-				*col = newCol
 			}
 		}
 	}
--- a/cmd/internal.go
+++ b/cmd/internal.go
@@ -178,19 +178,21 @@ func (r *ParseCommand) Run(assetPath string, lineage bool) error {
 			printErrorJSON(err)
 			return cli.Exit("", 1)
 		}
-		columnMetadata := make(map[string]map[string]string)
-		columnMetadata[asset.Name] = make(map[string]string, len(asset.Columns))
 
-		for _, col := range asset.Columns {
-			columnMetadata[asset.Name][col.Name] = col.Type
+		columnMetadata := make(sqlparser.Schema)
+
+		columnMetadata[asset.Name] = makeColumnMap(asset.Columns)
+		for _, upstream := range asset.Upstreams {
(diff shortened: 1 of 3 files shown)


cmd/internal.go
```diff
@@ -105,7 +115,7 @@ func (r *ParseCommand) ParsePipeline(assetPath string) error {
 
 	foundPipeline.WipeContentOfAssets()
 
-	js, err := json.Marshal(foundPipeline)
+	js, err := json.MarshalIndent(foundPipeline, "", "  ")
```
karakanb, line 118: I don't think we should do this, these are being passed onto the vscode extension, the smaller the payload the better

karakanb: you can always pipe this to `jq` to make it look nicer when working

cmd/internal.go
```diff
@@ -147,15 +157,47 @@ func (r *ParseCommand) Run(assetPath string) error {
 
 	foundPipeline.Assets = nil
 
-	js, err := json.Marshal(struct {
+	result := struct {
 		Asset    *pipeline.Asset    `json:"asset"`
+		Lineage  *sqlparser.Lineage `json:"lineage"`
```
karakanb, line 162: this is not what we have in mind, instead we should be using the lineage value to fill out the columns of the asset, `Asset.Columns` field

cmd/internal.go
```diff
@@ -208,3 +229,64 @@ func PatchAsset() *cli.Command {
...
+	if err != nil {
+		return fmt.Errorf("failed to parse column lineage: %w", err)
+	}
+
+	for _, lineageCol := range lineage.Columns {
+		for _, upstream := range lineageCol.Upstream {
+			upstreamAsset := pipe.GetAssetByName(upstream.Table)
+			if upstreamAsset == nil {
+				return fmt.Errorf("upstream asset not found: %s", upstream.Table)
+			}
```
karakanb, line 264: I don't think we should return errors, we should simply skip if we cannot find an upstream. we can add validation to this later on

cmd/internal.go
```diff
@@ -208,3 +229,64 @@ func PatchAsset() *cli.Command {
...
+
+			upstreamCol := upstreamAsset.GetColumnWithName(upstream.Column)
+			if upstreamCol == nil {
+				return fmt.Errorf("column %s not found in asset %s", upstream.Column, upstream.Table)
+			}
```
karakanb, line 269: same goes for this, maybe the upstream did not define the column, we shouldn't error here.

cmd/internal.go
```diff
@@ -208,3 +229,64 @@ func PatchAsset() *cli.Command {
...
+
+			newCol := *upstreamCol
+			newCol.Name = lineageCol.Name
+
+			if col := asset.GetColumnWithName(lineageCol.Name); col == nil {
+				asset.Columns = append(asset.Columns, newCol)
+			} else {
+				*col = newCol
+			}
```
karakanb, line 278: I think we should merge the whole object, not just the name.

the rules are:
- if the asset has defined any field, it wins
- if not, get it from the found upstream column

pkg/sqlparser/parser_test.go
```diff
@@ -213,6 +213,16 @@ func TestSqlParser_Lineage(t *testing.T) {
...
+		{
+			name: "yuvraj test",
+			sql: `
+				SELECT * from example2
+			`,
+			schema: Schema{
+				"example2": {"id": "str", "manager_id": "str"},
+			},
+			want: &Lineage{},
+		},
```
karakanb: don't forget to delete these

cmd/internal.go
y-bruin: that means we can remove else condition because column exist. Correct ?

albertobruin:
lgtm

PR #216, opened 2024-11-15
 added an error message for ingestr assets

Barbar1432:
Added an error message for hinting users should use asset.yml for ingestr assets

### DIFF
--- a/pkg/pipeline/pipeline.go
+++ b/pkg/pipeline/pipeline.go
@@ -1294,7 +1294,6 @@ func (b *Builder) CreateAssetFromFile(path string) (*Asset, error) {
 			// ends in yaml or yml but not asset.yml, task.yml etc
 			return nil, errors.New("You are trying to run a yaml file that doesn't end on: " + strings.Join(b.config.TasksFileSuffixes, ", "))
 		}
-
 	}
 
 	task, err := creator(path)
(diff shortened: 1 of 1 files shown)


pkg/pipeline/pipeline.go
```diff
@@ -1263,9 +1263,14 @@ func (b *Builder) CreateAssetFromFile(path string) (*Asset, error) {
...
-		creator = b.yamlTaskCreator
-		isSeparateDefinitionFile = true
+	if fileHasSuffix([]string{"yml", "yaml"}, path) {
+		if fileHasSuffix(b.config.TasksFileSuffixes, path) {
+			creator = b.yamlTaskCreator
+			isSeparateDefinitionFile = true
+		} else {
+			// ends in yaml or yml but not asset.yml, task.yml etc
+			return nil, errors.New("You are trying to run a yaml file that doesn't end on: " + strings.Join(b.config.TasksFileSuffixes, ", "))
+		}
```
karakanb, line 1273: try not to change the existing logic, that way we ensure we don't break the existing behavior anywhere:
```suggestion

	if fileHasSuffix(b.config.TasksFileSuffixes, path) {
		creator = b.yamlTaskCreator
		isSeparateDefinitionFile = true
	} else if fileHasSuffix([]string{".yml", ".yaml"}, path)
		// ends in yaml or yml but not asset.yml, task.yml etc
		return nil, errors.New("You are trying to run a yaml file that doesn't end on: " + strings.Join(b.config.TasksFileSuffixes, ", "))
	}
```

karakanb (changes requested):

karakanb (approved):

PR #217, opened 2024-11-15
Add new flag for column lineage in parse pipeline

y-bruin:
Examples: https://github.com/y-bruin/bruin-dev


cmd/internal.go
```diff
@@ -113,9 +122,15 @@ func (r *ParseCommand) ParsePipeline(assetPath string) error {
 		return cli.Exit("", 1)
 	}
 
-	foundPipeline.WipeContentOfAssets()
```
y-bruin, line 116: @albertobruin Do you know why we need `WipeContentOfAssets` ?

albertobruin: maybe because if we don't the size of the json would be too big?

cmd/internal.go
```diff
@@ -246,31 +264,36 @@ func ParseLineage(pipe *pipeline.Pipeline, asset *pipeline.Asset) error {
...
+					continue
+				}
+
+				if lineageCol.Name != "*" {
+					newCol := *upstreamCol
+					newCol.Name = lineageCol.Name
+					if col := asset.GetColumnWithName(lineageCol.Name); col == nil {
+						asset.Columns = append(asset.Columns, newCol)
+					}
+				}
```
karakanb, line 296: I'd invert these if statements:
```suggestion
			if upstream.Column == "*" {
				continue
			}
			
			if lineageCol.Name == "*" {
				continue
			}
				
			upstreamCol := upstreamAsset.GetColumnWithName(upstream.Column)
			if upstreamCol == nil {
				continue
			}
				
			newCol := *upstreamCol
			newCol.Name = lineageCol.Name
			if col := asset.GetColumnWithName(lineageCol.Name); col == nil {
				asset.Columns = append(asset.Columns, newCol)
			}
			
```

cmd/internal.go
```diff
@@ -286,3 +309,29 @@ func makeColumnMap(columns []pipeline.Column) map[string]string {
...
+			}
+		}
+
+		if err := ParseLineage(pipeline, asset); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
karakanb, line 337: I would have liked these to modify the assets within the `pipeline` object directly, otherwise they will all be lost since the pipeline is not being modified, am I wrong? 

in the end, I think we should design this to be a method like: `pipeline.PropagateColumns()` where it would fill out all the missing columns and modify itself or return a new pipeline object where all the columns are filled out. that way, we could use the same functionality outside of Bruin CLI as well.

cmd/internal.go
y-bruin: apart from quality checks is there anything we should ignore

karakanb: I don't think so.

albertobruin:
lgtm

PR #253, opened 2024-11-23
 add exclude tag  flag to run cmd

Barbar1432:
add exclude tag  flag to run cmd that excludes an asset from the pipeline execution


Barbar1432:
merge

cmd/run.go
```diff
@@ -124,6 +124,10 @@ func Run(isDebug *bool) *cli.Command {
 				Name:  "exp-use-winget-for-uv",
 				Usage: "use powershell to manage and install uv on windows, on non-windows systems this has no effect.",
 			},
+			&cli.StringFlag{
+				Name:  "exclude",
```
karakanb, line 128: I'd go with `--exclude-tags` and make it a string array, just like the `tags` one

cmd/run.go
```diff
@@ -756,3 +757,55 @@ func sendTelemetry(s *scheduler.Scheduler, c *cli.Context) {
...
+		// Mark assets with the given tag
+		s.MarkAll(scheduler.Succeeded)
+		s.MarkByTag(tag, scheduler.Pending, runDownstreamTasks)
+
+		infoPrinter.Printf("Running only the assets with tag '%s', found %d assets.\n", tag, len(assetsByTag))
+	} else {
+		assetsByTag = []*pipeline.Asset{}
+	}
+
+	return assetsByTag, nil
```
karakanb, line 810: I don't like this idea that this function is returning an irrelevant value that is used by another function, this seems like an optimization that could be done elsewhere. could we encapsulate this logic into another struct that's solely responsible from filtering what'll run?

cmd/run.go
Barbar1432: @karakanb As far as I see tag is not defined as a string array in the existing implementation --> `&cli.StringFlag{
				Name:    "tag",
				Aliases: []string{"t"},
				Usage:   "pick the assets with the given tag",
			},` would you like me to change it to string array ?

PR #256, opened 2024-11-25
Fix sql function parsing in column lineage

y-bruin:
Test Case: https://github.com/y-bruin/bruin-dev/blob/main/duckdb/assets/country.sql

Output: http://jsonblob.com/1310545255808032768

### DIFF
--- a/pkg/pipeline/lineage.go
+++ b/pkg/pipeline/lineage.go
@@ -124,6 +124,17 @@ func (p *LineageExtractor) processLineageColumns(asset *Asset, lineage *sqlparse
 			continue
 		}
 
+		if len(lineageCol.Upstream) == 0 {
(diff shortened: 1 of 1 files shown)


karakanb (approved):

pkg/pipeline/lineage.go
```diff
@@ -168,6 +179,7 @@ func (p *LineageExtractor) processLineageColumns(asset *Asset, lineage *sqlparse
 			if err := p.addColumnToAsset(asset, lineageCol.Name, upstreamAsset, upstreamCol); err != nil {
 				return err
 			}
+			continue
```
albertobruin, line 182: this continue is not necessary is it?

y-bruin: yes

PR #281, opened 2024-11-29
 add pipeline.yaml support

Barbar1432:
refactored Pipelinedefinition file to support pipeline.yaml files as well as all related functions to it

### DIFF
--- a/cmd/render.go
+++ b/cmd/render.go
@@ -98,7 +98,7 @@ func Render() *cli.Command {
 				}
 				return cli.Exit("", 1)
 			}
-			pipelinePath, err := path.GetPipelineRootFromTask(inputPath, pipelineDefinitionFile)
+			pipelinePath, err := path.GetPipelineRootFromTask(inputPath, pipelineDefinitionFiles)
 			if err != nil {
 				printError(err, c.String("output"), "Failed to get the pipeline path:")
 				return cli.Exit("", 1)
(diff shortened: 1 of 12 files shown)


cmd/render.go
```diff
@@ -298,3 +302,15 @@ func (r *RenderCommand) printErrorOrJSON(msg string) {
 func (r *RenderCommand) printErrorOrJsonf(msg string, args ...interface{}) {
 	r.printErrorOrJSON(fmt.Sprintf(msg, args...))
 }
+
+func getPipelineDefinitionFullPath(pipelinePath string) (string, error) {
```
albertobruin, line 306: shouldn't this return a string slice, like 
```
[]string{"/path/to/pipeline/pipeline.yml", "/path/to/pipeline/pipeline.yaml"} 
```

pkg/path/walk.go
```diff
@@ -27,43 +28,49 @@ func GetPipelinePaths(root, pipelineDefinitionFile string) ([]string, error) {
...
 	currentFolder := absoluteTaskPath
-	for currentFolder != filepath.VolumeName(currentFolder)+string(os.PathSeparator) && currentFolder != "/" {
-		tryPath := filepath.Join(currentFolder, pipelineDefinitionFile)
-		if _, err := os.Stat(tryPath); err == nil {
-			return currentFolder, nil
+	rootPath := filepath.VolumeName(currentFolder) + string(os.PathSeparator)
+	for currentFolder != rootPath && currentFolder != "/" {
+		for _, pipelineDefinition := range pipelineDefinitionFile {
+			tryPath := filepath.Join(currentFolder, pipelineDefinition)
+			if _, err := os.Stat(tryPath); err == nil {
```
albertobruin, line 65: I think `os.Stat(tryPath)` might not be the perfect way to check for existence. There could be a different error from "does not exist" 
See answer here https://stackoverflow.com/questions/12518876/how-to-check-if-a-file-exists-in-go

Barbar1432: Our existing implementation was using this line `os.Stat(tryPath)` that's why I kept it

cmd/render.go
```diff
@@ -298,3 +302,15 @@ func (r *RenderCommand) printErrorOrJSON(msg string) {
...
+	for _, pipelineDefinitionFile := range pipelineDefinitionFile {
```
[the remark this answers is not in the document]
Barbar1432: Thanks. I had first named **`PipelinedefinitionFile`** as **pipelineDefinitionFiles**  and afterwards I have decided it to rename it to PipelineDefinitionFile. I would change it to this if that's okay that would solve the issue below as well.
 `for _, pipelineDefinition := range pipelineDefinitionFile {`

karakanb (approved):

PR #283, opened 2024-11-29
Added support for conditions and join in lineage

y-bruin:


pkg/pipeline/pipeline.go
```diff
@@ -575,6 +575,8 @@ type Asset struct {
 
 	upstream   []*Asset
 	downstream []*Asset
+
+	CondtionalColumns []Column `json:"conditions_columns" yaml:"conditions_columns,omitempty" mapstructure:"conditions_columns"`
```
karakanb, line 579: we wouldn't want to store them here, intead let's store them inside the `Upstream` object, in the new `columns` field

pythonsrc/parser/main.py
```diff
@@ -3,6 +3,11 @@
 from sqlglot.optimizer import optimize, qualify
 from sqlglot.lineage import Node
 
+def extract_query(parsed):
+	cols = []
+	for expr in find_all_in_scope(parsed, exp.Column):
+		cols.append(expr)
+	return cols
```
karakanb, line 10: why don't we use/improve `extract_columns` instead of this new function? also, the name is not very clear, wdyt?

pythonsrc/parser/main.py
```diff
@@ -69,9 +74,21 @@ def get_column_lineage(query: str, schema: dict, dialect: str):
         return {"columns": []}
 
     result = []
+    conditions = []
+
+    conditional_cols = extract_query(optimized.find(exp.Where))
+    
+    
+	for col in conditional_cols:
+		col_name = str(str(col).split(".")[1]) if len(str(col).split(".")) == 2 else str(str(col).split(".")[0])
```
karakanb, line 83: what happens in different scenarios? e.g.
- `database.schema.table`
- `schema.table`
- `table`

PR #287, opened 2024-12-02
added filter-struct

Barbar1432:


y-bruin (approved):
It looks good to me.  consider moving the filter package to a separate directory. wdyt?

albertobruin (approved):

cmd/run.go
```diff
@@ -288,71 +288,20 @@ func Run(isDebug *bool) *cli.Command {
...
-				s.MarkAll(scheduler.Succeeded)
-				s.MarkAsset(task, scheduler.Pending, runDownstreamTasks)
-
-				if c.String("tag") != "" {
-					errorPrinter.Printf("You cannot use the '--tag' flag when running a single asset.\n")
-					return cli.Exit("", 1)
-				}
-			}
-
 			sendTelemetry(s, c)
```
albertobruin, line 300: Here we are sending the telemetry that gives us info on the assets that are going to be run *before* knowing what assets are going to be run. I think this needs to be moved after the filtering and marking

albertobruin (approved):

PR #313, opened 2024-12-12
Column lineage should return empty array for asset and column upstream

y-bruin:


albertobruin (approved):

albertobruin (approved):

pkg/pipeline/pipeline.go
```diff
@@ -508,7 +508,7 @@ type Upstream struct {
 	Type     string          `json:"type" yaml:"type" mapstructure:"type"`
 	Value    string          `json:"value" yaml:"value" mapstructure:"value"`
 	Metadata EmptyStringMap  `json:"metadata,omitempty" yaml:"metadata,omitempty" mapstructure:"metadata"`
-	Columns  []DependsColumn `json:"columns,omitempty" yaml:"columns,omitempty" mapstructure:"columns"`
+	Columns  []DependsColumn `json:"columns" yaml:"columns" mapstructure:"columns"`
```
karakanb, line 511: I think you should keep the `omitempty` for YAML, we wouldn't want to print empty stuff for yaml serialization

PR #327, opened 2024-12-17
Add validation rule

y-bruin:
- Introduced a new validation rule for detecting duplicate column checks within a single column and check if custom check have query or not
- Implemented the `ValidateDuplicateColumnChecks` and `ValidateCustomCheckQueryExists` function
- Updated the rules list in `list.go` to include the new rule with critical severity.

### DIFF
--- a/pkg/lint/list.go
+++ b/pkg/lint/list.go
@@ -154,14 +154,6 @@ func GetRules(fs afero.Fs, finder repoFinder, excludeWarnings bool, parser sqlPa
 			ApplicableLevels: []Level{LevelPipeline, LevelAsset},
 		},
 		&SimpleRule{
-			Identifier:       "duplicate-column-checks",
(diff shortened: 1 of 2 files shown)


pkg/lint/list.go
```diff
@@ -153,6 +153,22 @@ func GetRules(fs afero.Fs, finder repoFinder, excludeWarnings bool, parser sqlPa
 			AssetValidator:   ValidateDuplicateColumnNames,
 			ApplicableLevels: []Level{LevelPipeline, LevelAsset},
 		},
+		&SimpleRule{
+			Identifier:       "duplicate-column-checks",
```
y-bruin, line 157: @karakanb @albertobruin I think `duplicate-column-checks` is not required, wdyt ?

albertobruin (approved):

PR #341, opened 2024-12-19
Added a new flag `--continue` in run command 

y-bruin:


cmd/run.go
```diff
@@ -143,84 +146,51 @@ func Run(isDebug *bool) *cli.Command {
...
+				ExcludeTag:   c.String("exclude-tag"),
+				Only:         c.StringSlice("only"),
 			}
 
-			startDate, err := date.ParseTime(c.String("start-date"))
-			logger.Debug("given start date: ", startDate)
+			var startDate, endDate time.Time
+
+			var err error
+			startDate, endDate, inputPath, err := Validation(runConfig, c.Args().Get(0), logger)
```
karakanb, line 169: I'd give this function a more explicit name, sth like `ValidateRunConfig` or even put the function in the `RunConfig` struct maybe?

cmd/run.go
```diff
@@ -349,17 +328,7 @@ func Run(isDebug *bool) *cli.Command {
...
-				"endDate":     endDate.Format(time.RFC3339),
-				"runID":       runID,
-				"pipeline":    foundPipeline.Name,
-				"fullRefresh": strconv.FormatBool(c.Bool("full-refresh")),
-				"useUv":       strconv.FormatBool(c.Bool("use-uv")),
-				"tag":         c.String("tag"),
-				"excludeTag":  c.String("exclude-tag"),
-				"only":        strings.Join(c.StringSlice("only"), ","),
-			}, runID, statePath); err != nil {
+			if err := s.SavePipelineState(afero.NewOsFs(), *runConfig, runID, statePath); err != nil {
```
karakanb, line 331: why do you pass the runConfig by value instead of passing the pointer?

cmd/run.go
```diff
@@ -383,6 +352,144 @@ func Run(isDebug *bool) *cli.Command {
...
+		return nil, err
+	}
+
+	filter.IncludeTag = pipelineState.Parameters.Tag
+
+	filter.OnlyTaskTypes = pipelineState.Parameters.Only
+
+	filter.PushMetaData = pipelineState.Parameters.PushMetadata
+
+	filter.ExcludeTag = pipelineState.Parameters.ExcludeTag
```
karakanb, line 368: nitpick
```suggestion
	filter.IncludeTag = pipelineState.Parameters.Tag
	filter.OnlyTaskTypes = pipelineState.Parameters.Only
	filter.PushMetaData = pipelineState.Parameters.PushMetadata
	filter.ExcludeTag = pipelineState.Parameters.ExcludeTag
```

cmd/run.go
```diff
@@ -383,6 +352,144 @@ func Run(isDebug *bool) *cli.Command {
...
+
+	return pipelineState, nil
+}
+
+func GetPipeline(inputPath string, runConfig *scheduler.RunConfig, logger *zap.SugaredLogger) (*pipeline.Pipeline, bool, bool, *config.Config, error) {
```
karakanb, line 373: can you please simplify the return values from this function? it seems to be returning multiple bools, maybe we should be more explicit about what is being returned. maybe return a struct?

cmd/run.go
```diff
@@ -383,6 +352,144 @@ func Run(isDebug *bool) *cli.Command {
...
+		pipelinePath, err = path.GetPipelineRootFromTask(inputPath, pipelineDefinitionFiles)
+		if err != nil {
+			errorPrinter.Printf("Failed to find the pipeline this task belongs to: '%s'\n", inputPath)
+			return nil, runDownstreamTasks, runningForAnAsset, nil, err
+		}
+
+		if runConfig.Downstream {
+			infoPrinter.Println("The downstream tasks will be executed as well.")
+			runDownstreamTasks = true
+		}
```
karakanb, line 410: this logic could go outside of this function it seems like, since it is not very relevant to what's being done here and simply being returned.

PR #353, opened 2024-12-23
docs:  Continue from the last failed task

y-bruin:


cmd/run.go
```diff
@@ -120,7 +120,7 @@ func Run(isDebug *bool) *cli.Command {
 			},
 			&cli.BoolFlag{
 				Name:  "continue",
-				Usage: "Continue the pipeline from the last run",
+				Usage: "use continue to run the pipeline from the last failed task",
```
karakanb, line 123:
```suggestion
				Usage: "use continue to run the pipeline from the last failed asset",
```

docs/commands/run.md
```diff
@@ -29,6 +29,18 @@ bruin run [FLAGS] [optional path to the pipeline/asset]
 | `--push-metadata` | bool | `false` | Push metadata to the destination database if supported (currently BigQuery). |
 | `--tag` | str | - | Pick assets with the given tag. |
 | `--workers` | int | `16` | Number of workers to run tasks in parallel. |
+|  `--continue` | bool | `false` | Continue from the last failed task. |
```
karakanb, line 32:
```suggestion
|  `--continue` | bool | `false` | Continue from the last failed asset. |
```

docs/commands/run.md
```diff
@@ -29,6 +29,18 @@ bruin run [FLAGS] [optional path to the pipeline/asset]
...
+
+
+### Continue from the last failed task
```
karakanb, line 35:
```suggestion
### Continue from the last failed asset
```

docs/commands/run.md
```diff
@@ -29,6 +29,18 @@ bruin run [FLAGS] [optional path to the pipeline/asset]
...
+
+If you want to continue from the last failed task, you can use the `--continue` flag. This will run the pipeline/asset from the last failed task. Bruin will automatically retrive all the flags used in the last run. 
+
+```bash
+bruin run --continue 
+```
+
+NOTE: This will only work if the pipeline structure is not changed. If Pipeline/Asset dependencies are changed, you will need to run the pipeline/asset from the beginning. This is to ensure that the pipeline/asset is run in the correct order.
```
karakanb, line 43: We can format this a bit nicer:
```suggestion
> [!NOTE]
> This will only work if the pipeline structure is not changed. If the pipeline structure has changed in any way, including asset dependencies, you will need to run the pipeline/asset from the beginning. This is to ensure that the pipeline/asset is run in the correct order.
```

PR #364, opened 2024-12-30
Fix: use temp file for backup a assets in integration test

y-bruin:


integration-tests/integration-test.go
```diff
@@ -248,7 +248,12 @@ func customTestForContinue(currentFolder string) {
 		CompatibilityHash: "6a4a1598e729fea65eeaa889aa0602be3133a465bcdde84843ff02954497ff65",
 	})
 
-	if err := setupForContinue(currentFolder); err != nil {
+	tempfile, err := os.CreateTemp("", "tempfile-*.sql")
```
karakanb, line 251: I guess we should delete these files at the end of the test in an optimistic way, i.e. it's fine if the deletion fails as long as we can delete it most of the time.

PR #368, opened 2024-12-31
Seed data into the destination databases

y-bruin:
-  Introduced a new asset type for SQL, designed to facilitate loading data from CSV files into targeted databases.
   - `duckdb.seed`
   - `bq.seed`
- Added integration tests for testing  new assets type 
- Added one example in template and 
- Added docs


cmd/run.go
```diff
@@ -610,6 +610,15 @@ func setupExecutors(
 		}
 	}
 
+	if s.WillRunTaskOfType(pipeline.AssetTypeBigquerySeed) {
+		jinjaVariables := jinja.PythonEnvVariables(&startDate, &endDate, pipelineName, runID, fullRefresh)
+		if usePipForPython {
+			mainExecutors[pipeline.AssetTypeBigquerySeed][scheduler.TaskInstanceTypeMain] = python.NewLocalOperator(config, jinjaVariables)
```
karakanb, line 616: we should not be using the python operators for the seed, we need a new type of operator

pkg/pipeline/pipeline.go
```diff
@@ -1007,6 +1009,10 @@ func (p *Pipeline) GetAllConnectionNamesForAsset(asset *Asset) ([]string, error)
 		}
 
 		return []string{}, errors.Errorf("No default connection for type: '%s'", assetType)
+	} else if assetType == AssetTypeBigquerySeed {
+		// TODO: implement logic for seed
+		// Local CSV to Bigquery
+		return []string{"google_cloud_platform"}, nil
```
karakanb, line 1015: I don't think we need this

cmd/run.go
```diff
@@ -601,6 +601,11 @@ func setupExecutors(
 	// this should go away once we incorporate URIs into the assets
 	estimateCustomCheckType := s.FindMajorityOfTypes(pipeline.AssetTypeBigqueryQuery)
 
+	ingestrOperator, err := ingestr.NewSeedOperator(conn)
```
karakanb, line 604:
```suggestion
	seedOperator, err := ingestr.NewSeedOperator(conn)
```

docs/assets/sql-seed.md
```diff
@@ -0,0 +1,54 @@
+# SQL Seed Assets
```
karakanb, line 1: I wouldn't call it `SQL Seed Assets`, simply call them `Seed Assets`.
```suggestion
# Seed Assets
```

docs/assets/sql-seed.md
```diff
@@ -0,0 +1,54 @@
...
+Bruin supports running SQL assets against a variety of data platforms natively. Along with SQL assets, Bruin also supports running SQL Seed assets that will populate a table with data from a csv file. 
+
```
karakanb, line 3:
```suggestion
Seeds are CSV-files that contain data that is prepared outside of your pipeline that will be loaded into your data platform. Bruin supports seed assets natively, allowing you to simply drop a CSV file in your pipeline and ensuring the data is loaded to the destination platform accurately.

Seed assets are defined with `<platform>.seed` asset types, e.g.:
- for BigQuery: `bq.seed`
- for DuckDB: `duckdb.seed`
```

docs/assets/sql-seed.md
```diff
@@ -0,0 +1,54 @@
...
+You can define SQL Seed assets in a file ending with `.yaml`:
```
karakanb, line 4:
```suggestion
You can define seed assets in a file ending with `.yaml`:
```

karakanb: I think we should also add more details about the `path` parameter, since it is supposed to be optional.

docs/.vitepress/config.mjs
```diff
@@ -127,6 +127,7 @@ export default defineConfig({
                     {
                         text: "Asset Types", items: [
                             {text: "SQL", link: "/assets/sql"},
+                            {text: "SQL Seed", link: "/assets/sql-seed"},
```
karakanb, line 130:
```suggestion
                            {text: "Seed", link: "/assets/sql-seed"},
```

karakanb (approved):

PR #370, opened 2024-12-31
added goroutines for parallel execution

Barbar1432:


integration-tests/integration-test.go
```diff
@@ -128,38 +129,80 @@ func main() {
...
 	}
+	defer os.Remove(tempfile.Name())
 
 	workflows := getWorkflow(binary, currentFolder, tempfile.Name())
+	var wg sync.WaitGroup
+	errCh := make(chan error, len(workflows)) // Buffered channel to collect errors.
 
 	for _, workflow := range workflows {
-		err := workflow.Run()
+		wg.Add(1)
```
y-bruin, line 153: Let's execute the workflow without using goroutines, as creating temporary files and passing them is specific to this particular workflow and not a generic task.

y-bruin (approved):

PR #373, opened 2025-01-02
Fix lint

y-bruin:


Barbar1432:
maybe lets try `defer tempFile.Close()`   to  Ensure the file is closed after use  ?

pkg/connection/helper_test.go
```diff
@@ -10,7 +10,7 @@ func TestValidateServiceAccountFile(t *testing.T) {
 	t.Parallel()
 
 	// Create a temporary file for testing.
-	tempFile, err := os.CreateTemp("", "service_account.json")
+	tempFile, err := os.CreateTemp(t.TempDir(), "service_account.json")
```
Barbar1432, line 13: maybe lets try defer os.Remove(tempFile.Name()) before defer tempFile.Close() to Ensure the file is removed
after use ?

PR #376, opened 2025-01-02
handle cluster and partioning changes in Bq

Barbar1432:
Fixes the issue where changing cluster type causes the error: Cannot replace a table with a different partitioning spec. With this PR  if the full_refresh flag is passed, the system compares partitioning specs with the existing table. If they differ, it automatically drops the table, recreates it with the new partitioning specs


pkg/bigquery/db.go
```diff
@@ -31,6 +31,7 @@ type Selector interface {
 
 type MetadataUpdater interface {
```
albertobruin, line 32: I wouod change the name of the interface since now it's not just a metadata updated?

pkg/bigquery/db.go
```diff
@@ -271,3 +270,96 @@ func (d *Client) Ping(ctx context.Context) error {
 
 	return nil // Return nil if the query runs successfully
 }
+
+func (d *Client) DeleteTableIfPartitioningOrClusteringMismatch(ctx context.Context, tableName string, asset *pipeline.Asset) (err error) {
+	tableComponents := strings.Split(tableName, ".")
+	if len(tableComponents) != 2 {
```
sabrikaragonen, line 276: 3 levels should be allowed as well

pkg/bigquery/db.go
```diff
@@ -271,3 +270,96 @@ func (d *Client) Ping(ctx context.Context) error {
...
+		err = fmt.Errorf("table name must be in schema.table format, '%s' given", tableName)
+		return
+	}
+	tableRef := d.client.Dataset(tableComponents[0]).Table(tableComponents[1])
```
sabrikaragonen, line 280: I guess you're using it in both metadata and here, does it make sense to make a function for it?

albertobruin (approved):

PR #384, opened 2025-01-06
Make "path" an optional parameter for seed assets

y-bruin:


docs/assets/seed.md
```diff
@@ -16,7 +16,7 @@ You can see the "Data Platforms" on the left sidebar to see supported types.
 
 ## Parameters
 
-The `parameters` key in the configuration defines the parameters for the seed asset. The `path` parameter is the path to the CSV file that will be loaded into the data platform. path is relative to the asset definition file.
+The `parameters` key in the configuration defines the parameters for the seed asset. The `path` parameter is the path to the CSV file that will be loaded into the data platform. path is relative to the asset definition file. If the path is not provided, the asset name will be used to find the CSV file in the same directory as the asset definition file.
```
albertobruin, line 19: is this true? I thought for `countries.asset.yml` the file would be `prod.countries` or `prod.countries.csv`. what i read in the doc is not clear

albertobruin: I would be good to add an example though:

E.g For an asset with a name:

```
name: seed.raw_data
...
```

We would look for a `seed.raw_data.csv` in the same folder

PR #388, opened 2025-01-06
 create dataset if it doesnt exist

Barbar1432:
creates a bigquery dataset if it doesn't exist.  
Example  : Asset has the name  _new_dataset.product_categories_  but the user doesn't have the _new_dataset_  in bq , it will create it.


pkg/bigquery/db.go
```diff
@@ -395,3 +396,32 @@ func IsSameClustering(meta *bigquery.TableMetadata, asset *pipeline.Asset) bool
 
 	return true
 }
+
+func (d *Client) CreateDataSetIfNotExist(asset *pipeline.Asset, ctx context.Context) error {
+	tableName := asset.Name
+	tableComponents := strings.Split(tableName, ".")
+	var datasetName string
+	if len(tableComponents) == 2 {
```
albertobruin, line 404: is it possible to have a single (1) component/token?

pkg/bigquery/db.go
```diff
@@ -395,3 +396,32 @@ func IsSameClustering(meta *bigquery.TableMetadata, asset *pipeline.Asset) bool
...
+		datasetName = tableComponents[0]
+	} else if len(tableComponents) == 3 {
+		datasetName = tableComponents[1]
+	}
+	datasets := d.client.Datasets(ctx)
```
albertobruin, line 409: would it be possible to cache this in memory so in the case of for example having 2 BQ assets, we only make this call once? network calls are expensive :)

pkg/bigquery/db.go
sabrikaragonen: not in bigquery

albertobruin (approved):

albertobruin (approved):

PR #394, opened 2025-01-08
handle new databases in snowflake 

Barbar1432:
if user has new_database.table in  a snowflake asset which doesn't exists, create the database

### DIFF
--- a/pkg/snowflake/db.go
+++ b/pkg/snowflake/db.go
@@ -197,8 +197,8 @@ func (db *DB) SelectWithSchema(ctx context.Context, queryObj *query.Query) (*que
 	return result, nil
 }
 
-func (db *DB) CreateDataBaseIfNotExist(ctx context.Context, asset *pipeline.Asset) error {
-	databaseName := strings.Split(asset.Name, ".")[0]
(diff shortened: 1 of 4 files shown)


pkg/snowflake/db.go
```diff
@@ -190,3 +196,20 @@ func (db *DB) SelectWithSchema(ctx context.Context, queryObj *query.Query) (*que
 
 	return result, nil
 }
+
+func (db *DB) CreateDataBaseIfNotExist(ctx context.Context, asset *pipeline.Asset) error {
```
karakanb, line 200:
```suggestion
func (db *DB) CreateDatabaseIfNotExist(ctx context.Context, asset *pipeline.Asset) error {
```

pkg/snowflake/operator.go
```diff
@@ -140,6 +141,11 @@ func (o *QuerySensor) RunTask(ctx context.Context, p *pipeline.Pipeline, t *pipe
 		return err
 	}
 
+	err = conn.CreateDataBaseIfNotExist(ctx, t)
+	if err != nil {
+		return err
+	}
```
karakanb, line 147: I suggest doing this only if the asset name is two parts with `<schema>.<table>`

PR #397, opened 2025-01-09
Clickhouse

albertobruin:


pkg/clickhouse/operator.go
```diff
@@ -0,0 +1,80 @@
...
+	Render(task *pipeline.Asset, query string) (string, error)
+}
+
+type queryExtractor interface {
+	ExtractQueriesFromString(content string) ([]*query.Query, error)
+}
+
+type ClickHouseClient interface {
+	RunQueryWithoutResult(ctx context.Context, query *query.Query) error
+	Select(ctx context.Context, query *query.Query) ([][]interface{}, error)
```
Barbar1432, line 22: we can put SelectWithSchema and Ping here

pkg/clickhouse/operator.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return err
+	}
+
+	conn, err := o.connection.GetClickHouseConnection(connName)
+	if err != nil {
+		return err
+	}
+
+	for _, queryString := range materializedQueries {
```
Barbar1432, line 72: In another operators we seem to be running only one query per task , will this be different in click house ?

albertobruin: yes it will be. Some platforms allow you to send several statements in a single batch something like :

```sql
SELECT * from table;
INSERT ....;
DELETE ... FROM TABLE;
```

As long as they are separated by a `;`

Some others don't allow batches, you need to send statements one by one. 
Other examples of this besides Clickhouse are Athena or Synapse

PR #399, opened 2025-01-09
Snowflake  handle materialization type  changes

Barbar1432:
added :  user can change the materialization type after the table is already created

### DIFF
--- a/pkg/snowflake/db.go
+++ b/pkg/snowflake/db.go
@@ -226,11 +226,14 @@ func (db *DB) CreateSchemaIfNotExist(ctx context.Context, asset *pipeline.Asset)
 func (db *DB) HandleMaterializationTypeMismatch(ctx context.Context, asset *pipeline.Asset, errorMessage string, creationQuery *query.Query) error {
 	tableComponents := strings.Split(asset.Name, ".")
 	var schemaName string
+	var tableName string
 	switch len(tableComponents) {
 	case 2:
 		schemaName = strings.ToUpper(tableComponents[0])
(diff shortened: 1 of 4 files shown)


pkg/snowflake/operator.go
```diff
@@ -82,11 +84,24 @@ func (o BasicOperator) RunTask(ctx context.Context, p *pipeline.Pipeline, t *pip
 	if err != nil {
 		return err
 	}
+
 	err = conn.CreateSchemaIfNotExist(ctx, t)
 	if err != nil {
 		return err
 	}
-	return conn.RunQueryWithoutResult(ctx, q)
+	err = conn.RunQueryWithoutResult(ctx, q)
```
albertobruin, line 92: I would change this blocks to be:

```go
if err == nil {
    return nil
}
if o.materializer.IsFullRefresh() {
     if handleErr := ...
}

```

pkg/snowflake/db.go
```diff
@@ -222,3 +222,45 @@ func (db *DB) CreateSchemaIfNotExist(ctx context.Context, asset *pipeline.Asset)
...
+	case strings.Contains(errorMessage, "as VIEW"):
+		materializationType = "VIEW"
+
+	default:
+		return nil
+	}
+
+	dropQuery := query.Query{
+		Query: fmt.Sprintf("DROP %s IF EXISTS %s.%s", materializationType, schemaName, tableName),
+	}
```
karakanb, line 255: why do we do this based on the error instead of comparing the type in the destination and the type in our query, without waiting for an error?

PR #400, opened 2025-01-10
Snowflake column description push

Barbar1432:


pkg/snowflake/db.go
```diff
@@ -282,3 +282,79 @@ func (db *DB) RecreateTableOnMaterializationTypeMismatch(ctx context.Context, as
...
+	for _, col := range asset.Columns {
+		colsByName[col.Name] = &col
+		if col.Description != "" {
+			anyColumnHasDescription = true
+		}
+	}
+
+	if asset.Description == "" && (len(asset.Columns) == 0 || !anyColumnHasDescription) {
+		return errors.New("no metadata to push: table and columns have no descriptions")
+	}
```
karakanb, line 312: I think the `anyColumnHasDescription` is not necessary anymore, given that we've put this behind a flag anyway.
```suggestion
	colsByName := make(map[string]*pipeline.Column, len(asset.Columns))
	for _, col := range asset.Columns {
		colsByName[col.Name] = &col
	}

	if asset.Description == "" && (len(asset.Columns) == 0) {
		return errors.New("no metadata to push: table and columns have no descriptions")
	}
```

pkg/snowflake/db.go
```diff
@@ -282,3 +282,79 @@ func (db *DB) RecreateTableOnMaterializationTypeMismatch(ctx context.Context, as
...
+		comment := ""
+		if row[1] != nil {
+			comment = row[1].(string)
+		}
+		existingComments[columnName] = comment
+	}
+
+	// Find columns that need updates
+	for _, col := range asset.Columns {
+		if col.Description != "" && existingComments[col.Name] != col.Description {
```
karakanb, line 337: this seems like it would add comments only to the columns that already have existing comments, which means it wouldn't add new comments to cols that didn't have them before, no? 
```suggestion
		if existingComments[col.Name] != col.Description {
```

pkg/snowflake/db.go
```diff
@@ -282,3 +282,79 @@ func (db *DB) RecreateTableOnMaterializationTypeMismatch(ctx context.Context, as
...
+			updateQuery := fmt.Sprintf(
+				`ALTER TABLE %s.%s.%s MODIFY COLUMN %s COMMENT '%s'`,
+				db.config.Database, schemaName, tableName, col.Name, col.Description,
+			)
+			if err := db.RunQueryWithoutResult(ctx, &query.Query{Query: updateQuery}); err != nil {
+				return errors.Wrapf(err, "failed to update description for column %s", col.Name)
+			}
```
karakanb, line 344: maybe we should join multiple queries and run all of them at once? does it work?

pkg/snowflake/operator.go
```diff
@@ -94,7 +95,19 @@ func (o BasicOperator) RunTask(ctx context.Context, p *pipeline.Pipeline, t *pip
...
+		return err
+	}
+
+	if p.MetadataPush.HasAnyEnabled() {
+		err = conn.PushColumnDescriptions(ctx, t)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
karakanb, line 110: we shouldn't do it here, instead this should be a new operator that is executed outside of the main asset, otherwise we'll be rerunning the query when the user wants to only push the metadata. take a look at the bigquery implementation for an inspiration.

pkg/snowflake/db.go
Barbar1432: col.description is the , description in the asset not the description in the snowflake

pkg/snowflake/db.go
```diff
@@ -282,3 +282,74 @@ func (db *DB) RecreateTableOnMaterializationTypeMismatch(ctx context.Context, as
...
+	for _, col := range asset.Columns {
+		if col.Description != "" && existingComments[col.Name] != col.Description {
+			query := fmt.Sprintf(
+				`ALTER TABLE %s.%s.%s MODIFY COLUMN %s COMMENT '%s'`,
+				db.config.Database, schemaName, tableName, col.Name, col.Description,
+			)
+			updateQueries = append(updateQueries, query)
+		}
+	}
+	if len(updateQueries) > 0 {
```
Barbar1432, line 337: I am batching all of the queries here to push the column descriptions at the same time , should I do this in smaller batches , if so how can I decide the size of the batches?

pkg/snowflake/db.go
```diff
@@ -282,3 +282,74 @@ func (db *DB) RecreateTableOnMaterializationTypeMismatch(ctx context.Context, as
...
+		batchQuery := strings.Join(updateQueries, "; ")
```
Barbar1432, line 338: I am batching all of the queries here to push the column descriptions at the same query , should I do this in smaller batches , if so how can I decide the size of the batches?

albertobruin (approved):

PR #404, opened 2025-01-15
Hotfix: Remove lint from run command

y-bruin:


cmd/run.go
```diff
@@ -209,15 +209,6 @@ func Run(isDebug *bool) *cli.Command {
 				infoPrinter.Printf("Running only the asset '%s'\n", task.Name)
 			}
 
-			parser, err := sqlparser.NewSQLParser(false)
-			if err != nil {
-				printError(err, runConfig.Output, "Could not initialize sql parser")
-			}
-
-			if err := CheckLint(parser, pipelineInfo.Pipeline, inputPath, logger); err != nil {
```
karakanb, line 217: I think we misunderstood each other: we want to keep the linting, but not the sql parsing stuff.

PR #406, opened 2025-01-15
cmd: run: add --debug-ingestr-src flag

turtleDev:
This flag allows developers to provide the path of an `ingestr` source tree that they would like to use, instead of the version bruin CLI was compiled with.

This should make it easy for devs to test integrations in bruin CLI without having to create new releases

**Example**
```
bruin run --debug-ingestr-src /home/user/ingestr pipeline/assets/download.asset.yml
```

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -141,6 +141,10 @@ func Run(isDebug *bool) *cli.Command {
 				Name:  "exp-use-winget-for-uv",
 				Usage: "use powershell to manage and install uv on windows, on non-windows systems this has no effect.",
 			},
+			&cli.StringFlag{
+				Name:  "debug-ingestr-src",
+				Usage: "Use ingestr from the given path instead of the builtin version.",
+			},
 		},
 		Action: func(c *cli.Context) error {
 			defer func() {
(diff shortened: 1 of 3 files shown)


turtleDev:
There's a small edge case with this change that I need to verify.

when using a local ingestr at a given tag, say `x.y.z`, I'm not sure if it will update it when run without the debug flag when the latest version is also `x.y.z`. 

Consider this PR on hold until I confirm this.

turtleDev:
I've confirmed that this bug exists.

turtleDev:
Fixed. We do a forced `--reinstall` of the uv tool on each invocation. This adds 300-500ms to runtime, but is marginal when compared to total runtime of a typical job.

This wasn't my first choice. I expored altering uv `site-packages` directory to change the version of a locally installed version of ingestr to `vX.Y.Z` to `vX.Y.Z-dev`. Which _kind of_ worked, but caused it's own problems. It would cause errors when we tried to reinstall it next time. It would require precise (and hacky) manipulation of the environment managed by uv. This approach was also fragile against any future changes in uv.

In light of these challenges, the former approach seemed robust and simple.

cmd/run.go
```diff
@@ -344,6 +348,7 @@ func Run(isDebug *bool) *cli.Command {
 			runCtx = context.WithValue(runCtx, pipeline.RunConfigEndDate, endDate)
 			runCtx = context.WithValue(runCtx, executor.KeyIsDebug, isDebug)
 			runCtx = context.WithValue(runCtx, python.CtxUseWingetForUv, runConfig.ExpUseWingetForUv) //nolint:staticcheck
+			runCtx = context.WithValue(runCtx, python.LocalIngestr, c.String("debug-ingestr-src"))
```
y-bruin, line 351: please also add it in run config https://github.com/bruin-data/bruin/pull/406/files#diff-8146f8148ccbf6711d65f532f6ab9a7c8dfbdc3960c7ffc974f17d6d224dd349L157

PR #411, opened 2025-01-17
add docs

albertobruin:


docs/platforms/athena.md
```diff
@@ -48,6 +47,7 @@ select
     avg(session_duration) as avg_session_duration
 from raw_web_traffic
 group by date;
+@bruin */
```
karakanb, line 50: seems like a copy-paste mistake, this should have been before the query.

docs/platforms/athena.md
```diff
@@ -66,6 +65,7 @@ select
     avg(total_amount) as avg_order_value
 from sales_data
 group by order_date;
+@bruin */
```
karakanb, line 68: same here.

docs/platforms/athena.md
```diff
@@ -85,6 +84,7 @@ select
...
```
karakanb, line 87: same here.

PR #412, opened 2025-01-17
Default values for assets

y-bruin:


pkg/pipeline/pipeline.go
```diff
@@ -1412,5 +1417,34 @@ func (b *Builder) CreateAssetFromFile(path string) (*Asset, error) {
...
+
+		if len(task.Secrets) == 0 {
+			task.Secrets = []SecretMapping{}
+		}
+
+		existingSecrets := make(map[string]bool)
+		for _, secret := range task.Secrets {
+			existingSecrets[secret.SecretKey] = true
+		}
+		for _, secret := range foundPipeline.DefaultValues.Secrets {
```
Barbar1432, line 1443: In the documentation , we say . _If an asset has its own specified value, it will take precedence over the default._  but here we seem to be adding the secrets on top of the existing ones , should we write a more clear description ?

y-bruin: Yes you are correct

pkg/pipeline/pipeline.go
```diff
@@ -1384,34 +1377,78 @@ func fileHasSuffix(arr []string, str string) bool {
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "error converting default values to bytes")
+		}
+		var result taskDefinition
+
+		err = path.ConvertYamlToObject(defaultValuesBytes, &result)
+		if err != nil {
+			return nil, err
+		}
+
```
karakanb, line 1426: I don't think we should do this, this would allow people to override everything by default, e.g. things like `depends` or so, we wouldn't want that.

pkg/pipeline/pipeline.go
```diff
@@ -1384,34 +1377,78 @@ func fileHasSuffix(arr []string, str string) bool {
...
+	task.DefinitionFile.Name = filepath.Base(filePath)
+	task.DefinitionFile.Path = filePath
 	task.DefinitionFile.Type = CommentTask
 	if isSeparateDefinitionFile {
 		task.DefinitionFile.Type = YamlTask
 	}
 
+	if foundPipeline != nil && foundPipeline.DefaultValues != nil {
+		defaultValuesBytes, err := json.Marshal(foundPipeline.DefaultValues)
```
karakanb, line 1416: you can use https://github.com/go-viper/mapstructure/ for this instead of serializing/deserializing

karakanb (approved):

PR #426, opened 2025-01-27
Refactor connections test

Barbar1432:
changes made : 

- made environment flag optional  --> uses the default env if the flag isn't passed 
-  deleted the  type flag --> realized this was only used for error messages creates a worst user experience

### DIFF
--- a/cmd/connections.go
+++ b/cmd/connections.go
@@ -349,15 +349,9 @@ func PingConnection() *cli.Command {
 		Usage: "Test the validity of a connection in an environment",
 		Flags: []cli.Flag{
 			&cli.StringFlag{
-				Name:     "environment",
-				Aliases:  []string{"e", "env"},
-				Usage:    "the name of the environment (e.g., dev, prod)",
-				Required: true,
-			},
-			&cli.StringFlag{
-				Name:     "type",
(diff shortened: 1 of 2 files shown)


cmd/connections.go
```diff
@@ -391,13 +384,20 @@ func PingConnection() *cli.Command {
...
+				if err != nil {
+					printErrorForOutput(output, errors2.Wrap(err, "failed to select the environment"))
+					return cli.Exit("", 1)
+				}
+			} else {
+				err = cm.SelectEnvironment(environment)
+				if err != nil {
+					printErrorForOutput(output, errors2.Wrap(err, "failed to select the environment"))
+					return cli.Exit("", 1)
+				}
```
karakanb, line 398: rule of thumb: if you need an `else` there's prolly a cleaner way to do it.
```suggestion
			if environment == "" {
				environment = cm.DefaultEnvironmentName
			}	
			err = cm.SelectEnvironment(environment)
			if err != nil {
				printErrorForOutput(output, errors2.Wrap(err, "failed to select the environment"))
				return cli.Exit("", 1)
			
```

karakanb (approved):

PR #429, opened 2025-01-28
[WIP] add clickhouse connection

sanjushahgupta:


docs/platforms/clickhouse.md
```diff
@@ -11,12 +11,25 @@ connections:
         - name: "connection_name"
           username: "clickhouse"
           password: "XXXXXXXXXX"
-          host: "some-clickhouse-host.somedomain.com"
+          host: "some-clickhouse-host.somedomain.com"   
           port: 9000
-          database: "dev"
+          database: "dev" #Optional for other assets, but should not be provided when using ClickHouse as an ingestr destination, as ingestr takes the database name from the asset file.
```
albertobruin, line 16: I would change 
`should not be provided when using ClickHouse as an ingestr destination`
to 
`uneffective when using ClickHouse as an ingestr destination`

docs/platforms/clickhouse.md
```diff
@@ -11,12 +11,25 @@ connections:
         - name: "connection_name"
           username: "clickhouse"
           password: "XXXXXXXXXX"
-          host: "some-clickhouse-host.somedomain.com"
+          host: "some-clickhouse-host.somedomain.com"   
           port: 9000
-          database: "dev"
+          database: "dev" #Optional for other assets, but should not be provided when using ClickHouse as an ingestr destination, as ingestr takes the database name from the asset file. 
+          http_port: 8123 #Only specify if you are using clickhouse as ingestr destination, by default it is 8123
```
albertobruin, line 17: `#Only specify if you are using clickhouse as ingestr destination, by default it is 8123`
to
`#Only needed if you are using clickhouse as ingestr destination` (it doesn't have any default as you can see here https://github.com/bruin-data/bruin/pull/429/files#diff-f410faaf5345121c53b2dcf8b98322f965dc07410564d810b6069a96d4aed260R350)

PR #432, opened 2025-01-30
Refactor init cmd 

terzioglub:
**Old Process:**
For Every Initialized Template:

A separate _.bruin.yml_ file was created inside the template, even if one already existed elsewhere.

**New Process:**

task: running bruin init with template X

if current folder is git repo:
    if .bruin.yml exist in current folder:
        merge .bruin.yml from current folder with the one from template X
    else:
        create .bruin.yml in current folder from contents of .bruin.yml of template X
        create folder X in current folder with pipeline contentx except the .bruin.yml
else (current foder is not git repo):
    
    go up the folder tree structure until we find a git repo or we reach the very root of the file system
    
    if we reach the very root of the filesystem:
        create "bruin" folder
        init git repo in "bruin" folder
        create .bruin.yml based on the template X and place bruin folder
        create folder X inside bruin folder

    else (if we reach an existing git repo in the ancestors in folder Y):
        do we also find a .bruin.yml:
            merge .bruin.yml of pipeline X with the one in folder Y
            create folder X with contents of the pipeline except for .bruin.yml (since we mertged)
        else:
            create .bruin.yml in folder Y based on the one from pipeline X
            create folder X with contents of the pipeline except for .bruin.yml 

[asciicast](https://asciinema.org/a/6plpocOVlG4BTBedxFoIUzYeu)

Note: If `.bruin.yaml` is not available in template then we don't do anything. If `.bruin.yaml` exist then we merge the config


cmd/init.go
```diff
@@ -78,6 +78,47 @@ func (m model) View() string {
...
+	for templateEnvName, templateEnv := range templateConfig.Environments {
+		if err := mergeEnvironment(centralConfig, templateEnvName, templateEnv); err != nil {
+			return fmt.Errorf("failed to merge environment %s: %w", templateEnvName, err)
+		}
+	}
+
+	return nil
+}
+
+func mergeEnvironment(centralConfig *config.Config, templateEnvName string, templateEnv config.Environment) error {
```
karakanb, line 102: let's say the user has two envs, `staging` and `prod`, `staging` being marked as the `default`. if we have a `.bruin.yml` in our template and that has an env called `default`, it seems like we would create it as a new environment, correct? in that case I suggest using the user's default environment and merge into that.

karakanb (commented):
two notes:
- let's add an integration test that covers the changes here.
- make sure we are not overriding if there are existing connections with the same name.

cmd/init.go
```diff
@@ -288,7 +288,6 @@ func Init() *cli.Command {
 
 			successPrinter.Printf("\n\nA new '%s' pipeline created successfully in folder '%s'.\n", templateName, inputPath)
 			infoPrinter.Println("\nYou can run the following commands to get started:")
-			infoPrinter.Printf("\n    cd %s\n", inputPath)
```
terzioglub, line 210: I think it makes sense to delete this. I think it might be good to add here
infoPrinter.Printf(" bruin run path\n\n") path is either just the pipeline name or bruin/pipeline-name

albertobruin (approved):

PR #436, opened 2025-02-02
ClickHouse: Add Support for Secure Connection

sanjushahgupta:
Following changes are made in this PR:
- Improved connection handling by allowing users to specify secure or non-secure connections using the secure parameter
- By default, the secure parameter is set to 1, i.e secure connection.

Screenshot 2025-02-03 at 11 15 14

Screenshot 2025-02-03 at 11 15 38

### DIFF
--- a/pkg/clickhouse/config.go
+++ b/pkg/clickhouse/config.go
@@ -2,6 +2,8 @@ package clickhouse
 
 import (
 	"fmt"
+	"net/url"
+	"strconv"
 
 	click_house "github.com/ClickHouse/clickhouse-go/v2"
 )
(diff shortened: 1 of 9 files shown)


pkg/clickhouse/config.go
```diff
@@ -28,10 +29,18 @@ func (c *Config) ToClickHouseOptions() *click_house.Options {
 }
 
 func (c *Config) GetIngestrURI() string {
+	//nolint:nosprintfhostport
+	uri := fmt.Sprintf("clickhouse://%s:%s@%s:%d", c.Username, c.Password, c.Host, c.Port)
```
turtleDev, line 33: I would recommend using `url` package to build the URI. It will automatically escape URI characters, making this a lot more robust.

turtleDev: Here's an example:
https://github.com/bruin-data/bruin/blob/main/pkg%2Fgoogleads%2Fconfig.go#L16

turtleDev (changes requested):

pkg/clickhouse/config.go
```diff
@@ -13,6 +13,7 @@ type Config struct {
 	Port     int
 	Database string
 	HTTPPort int
+	Secure   *int
```
turtleDev, line 16: Can we avoid using pointers? 

Instead treat `0` as the null value.

turtleDev (approved):

PR #445, opened 2025-02-06
Patch/lineage column refactor

y-bruin:
- Refactor lineage column processing for modularity
- Fix column name in asset upstream if sql query use `as`  for rename the column name


pkg/pipeline/lineage.go
```diff
@@ -110,63 +110,75 @@ func (p *LineageExtractor) parseLineage(foundPipeline *Pipeline, asset *Asset, m
 	return p.processLineageColumns(foundPipeline, asset, lineage)
 }
 
-func (p *LineageExtractor) processLineageColumns(foundPipeline *Pipeline, asset *Asset, lineage *sqlparser.Lineage) error {
-	if lineage == nil {
-		return nil
-	}
+func (p *LineageExtractor) mergeAstricColumns(foundPipeline *Pipeline, asset *Asset, lineageCol sqlparser.ColumnLineage) error {
```
albertobruin, line 113: I think there's a typo here? Should it be Asterisk?

y-bruin: yes

albertobruin (approved):

PR #452, opened 2025-02-13
feat: added extends to asset spec

y-bruin:
Changes:
- Split the logic for reading asset defination and mutating the default logic into two seprate function 
- Added a new key in asset defination i.e. `extends`, Please read the docs [here](https://github.com/bruin-data/bruin/pull/452/files#diff-c5c035ef18fec4d7765e6e3678e34ca9b91084e86681a039d784a974841a50a1L141) 
- Added docs and Integration test  for this feature 
- Added Integration Test for `extends` available [here](https://github.com/bruin-data/bruin/blob/2edc5dc5c56caa91eea9b2a025a521e6e4ba0f8a/integration-tests/test-pipelines/parse-asset-extends/pipeline.yml)


integration-tests/test-pipelines/parse-asset-extends/assets/country.sql
```diff
@@ -0,0 +1,14 @@
+/* @bruin
+name: country
+type: duckdb.sql
+materialization:
+   type: table
+
+extends: 
+  - Customer
+
```
karakanb, line 9: what if we had some other columns, and some other extends maybe?
```suggestion
extends: 
  - Customer
 
columns:
  - name: mycol1
  - name: street_name
    extends: Address.Street
```

pkg/pipeline/testdata/pipeline/first-pipeline_windows.json
```diff
@@ -22,6 +22,7 @@
       "description": "This is a hello world task",
       "type": "bash",
       "tags": [],
+      "extends": null,
```
karakanb, line 25: an empty array by default would be better.

pkg/pipeline/pipeline.go
```diff
@@ -1431,6 +1432,29 @@ func (b *Builder) CreateAssetFromFile(filePath string, foundPipeline *Pipeline)
...
+						task.Columns = append(task.Columns, Column{
+							EntityAttribute: &EntityAttribute{
+								Entity:    entity.Name,
+								Attribute: attribute.Name,
+							},
+						})
+					}
+					break
+				}
+			}
```
karakanb, line 1454:
```suggestion
			for _, entity := range entities {
				if entity.Name != extend {
				    continue
				}
				
				for _, attribute := range entity.Attributes {
					task.Columns = append(task.Columns, Column{
						EntityAttribute: &EntityAttribute{
							Entity:    entity.Name,
							Attribute: attribute.Name,
						},
					})
				}
				
				break
			}
```

karakanb (commented):
check what happens with the output of `bruin format` as well please

integration-tests/test-pipelines/format-if-changed-pipeline/assets/correctly-formatted.sql
```diff
@@ -5,6 +5,7 @@ type: duckdb.sql
 
 materialization:
   type: table
+extends: []
```
karakanb, line 8: we wouldn't want to show it if it is an empty list.

pkg/lint/lint_test.go
```diff
@@ -128,11 +128,11 @@ func TestLinter_Lint(t *testing.T) {
 					return []string{"path/to/pipeline1", "path/to/pipeline2", "path/to/pipeline2_some_other_name"}, nil
 				},
 				setupBuilderMock: func(m *mockPipelineBuilder) {
-					m.On("CreatePipelineFromPath", "path/to/pipeline1").
+					m.On("CreatePipelineFromPath", "path/to/pipeline1", mock.Anything).
```
albertobruin, line 131: why not use the actual boolean value insteead of `mock.Anything`?

y-bruin: for lint it is always true

karakanb (dismissed):

albertobruin (approved):

albertobruin (approved):

PR #461, opened 2025-02-20
jinja: add_month filter 

terzioglub:


docs/assets/templating/filters.md
```diff
@@ -14,8 +14,25 @@ Bruin CLI supports various default filters supported by Jinja, you can see the [
...
+
+```jinja
+{{ end_datetime | add_years(2) }}
+```
+
 Bruin supports various date-related filters to format dates that can be chained together.
 
+### `add_months`
+
+Adds the given number of months to the date.
```
karakanb, line 29: I think you should add a bit more details on how is this different than just adding 30 days

karakanb (approved):

PR #464, opened 2025-02-20
Refactor query cmd

terzioglub:
Three new flags added  :**limit** --> to limit the rows returned 
**asset** --> asset where we get the query from
**env** --> env the asset is being used on

Previously we only had the option to use **connection** and **query** flags now we can either use these two in tandem or just pass the  **asset**  if its in the default environment or combine with the **env** if the asset is not used in the default environment.  This second option detects the connection that the asset needs and unmaterializes  the query from the asset and then runs the query

### DIFF
--- a/cmd/fetch.go
+++ b/cmd/fetch.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	path2 "path"
 	"path/filepath"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/bruin-data/bruin/pkg/config"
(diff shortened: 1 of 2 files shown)


cmd/fetch.go
```diff
@@ -29,95 +40,232 @@ func Query() *cli.Command {
...
+		if err != nil {
+			return nil, "", errors.Wrapf(err, "failed to use the environment '%s'", env)
+		}
+	}
+
+	// Create extractor with jinja renderer
+	startDate := time.Now() // You might want to make these configurable
+	endDate := time.Now()
+	extractor := &query.WholeFileExtractor{
+		Fs:       afero.NewOsFs(),
```
albertobruin, line 181: let's start a single of this

albertobruin (approved):

PR #479, opened 2025-02-28
Sf time materialization

terzioglub:


pkg/snowflake/operator.go
```diff
@@ -77,6 +77,18 @@ func (o BasicOperator) RunTask(ctx context.Context, p *pipeline.Pipeline, t *pip
...
+		if err != nil {
+			return errors.Wrap(err, "cannot re-extract/render materialized query for time_interval strategy")
+		}
+
+		if len(renderedQueries) == 0 {
+			return errors.New("rendered queries unexpectedly empty")
+		}
+
+		q.Query = renderedQueries[0].Query
+	}
```
karakanb, line 91: what's the purpose of this section? why do we re-extract the queries? if that's due to jinja rendering I'd suggest reorganizing this code a bit more to make it work in all cases properly

terzioglub: time_interval strategy adds jinja templates to the query , that's why it needs some form of reprocessing but I will look for a cleaner way to do this

karakanb (approved):

PR #489, opened 2025-03-04
allow custom query parameter for mssql

nicokant:
Allow to override the default query parameter for the database connection, this should allow to use a different driver, as well as different authentication options (`Trusted_Connection=yes` for example).

I'm not sure if this is enough, feel free to give me directions or to modify the code :)


nicokant:
Fixed the liniting and added some testing, let me know if there is something else I can do :)

nicokant:
@karakanb could I ask you if it would be possible for you to review and (if it's alright) merge these changes? I depend on this to complete some data migration tasks and without it won't be possible to even connect to my SQL Server instance :)

Thank you very much in advance for your work!

y-bruin (approved):

PR #490, opened 2025-03-04
add kinesis connection 

sanjushahgupta:


pkg/config/connections.go
```diff
@@ -547,3 +547,14 @@ type GCSConnection struct {
 func (c GCSConnection) GetName() string {
 	return c.Name
 }
+
+type KinesisConnection struct {
+	Name               string `yaml:"name,omitempty" json:"name" mapstructure:"name"`
+	AWSACCESSKEYID     string `yaml:"aws_access_key_id,omitempty" json:"aws_access_key_id" mapstructure:"aws_access_key_id"`
+	AWSSECRETACCESSKEY string `yaml:"aws_secret_access_key,omitempty" json:"aws_secret_access_key" mapstructure:"aws_secret_access_key"`
```
turtleDev, line 554:
```suggestion
	AccessKeyID     string `yaml:"access_key_id,omitempty" json:"access_key_id" mapstructure:"access_key_id"`
	SecretAccessKey string `yaml:"secret_access_key,omitempty" json:"secret_access_key" mapstructure:"secret_access_key"`
```

pkg/kinesis/config.go
```diff
@@ -0,0 +1,13 @@
...
+import "fmt"
+
+type Config struct {
+	AWSAccessKeyID     string
+	AWSSecretAccessKey string
+	Region             string
+}
+
+func (c *Config) GetIngestrURI() string {
+	return fmt.Sprintf("kinesis://?aws_access_key_id=%s&aws_secret_access_key=%s&region_name=%s", c.AWSAccessKeyID, c.AWSSecretAccessKey, c.Region)
```
turtleDev, line 12: use `url` package to build the URI. 

This is unsafe, because access key id/secret access key may contain URI delimiters (for instance `=`,`&` etc).

docs/ingestion/kinesis.md
```diff
@@ -0,0 +1,55 @@
...
+```yaml
+name: public.kinesis
+type: ingestr
+connection: postgres
+
+parameters:
+  source_connection: kinesis
+  source_table: 'test_stream'
+
+  destination: postgres
```
albertobruin, line 39: you don't explain what is the `destination` in the doc. I think we should. Also is it like this in other platforms' documentation?

sanjushahgupta: Yes, it is like this in other platform's documentation. I will add it.

pkg/config/connections.go
```diff
@@ -547,3 +547,14 @@ type GCSConnection struct {
 func (c GCSConnection) GetName() string {
 	return c.Name
 }
+
+type KinesisConnection struct {
+	Name            string `yaml:"name,omitempty" json:"name" mapstructure:"name"`
+	ACCESSKEYID     string `yaml:"access_key_id,omitempty" json:"access_key_id" mapstructure:"access_key_id"`
```
turtleDev, line 553: What's with the all caps? Why not title case? `AccessKeyID` ?

turtleDev (approved):

PR #516, opened 2025-03-12
patch: Lineage standard error message for user

y-bruin:


pkg/lineage/lineage.go
```diff
@@ -91,54 +91,72 @@ func (p *LineageExtractor) ColumnLineage(foundPipeline *pipeline.Pipeline, asset
...
 
 	lineage, err := p.sqlParser.ColumnLineage(query, dialect, metadata)
 	if err != nil {
-		return fmt.Errorf("failed to parse column lineage: %w", err)
+		return nil, fmt.Errorf("failed to parse lineage: %w", err)
 	}
 
 	if len(lineage.Errors) > 0 {
-		return fmt.Errorf("failed to parse column lineage: %s", strings.Join(lineage.Errors, ", "))
+		fmt.Println("lineage", lineage.Errors)
```
karakanb, line 145: don't forget to remove this

pkg/lineage/lineage.go
```diff
@@ -91,54 +91,72 @@ func (p *LineageExtractor) ColumnLineage(foundPipeline *pipeline.Pipeline, asset
...
+		return &LineageIssue{
+			Task:        asset,
+			Description: strings.Join(lineage.Errors, ", "),
+			Context: []string{
+				asset.ExecutableFile.Content,
+			},
+		}, fmt.Errorf("failed to parse lineage: %s", strings.Join(lineage.Errors, ", "))
```
karakanb, line 152: I don't think we should return extra messages here, since these are lineage-specific errors
```suggestion
		}, fmt.Errorf("%s", strings.Join(lineage.Errors, ", "))
```

y-bruin: It's ok because user don't care about error here

karakanb (approved):

PR #521, opened 2025-03-17
Time interval mat databricks

terzioglub:


pkg/databricks/materialization.go
```diff
@@ -131,3 +132,32 @@ func buildCreateReplaceQuery(task *pipeline.Asset, query string) ([]string, erro
 		fmt.Sprintf(`ALTER TABLE %s RENAME TO %s;`, tempTableName, task.Name),
 	}, nil
 }
+
+func buildTimeIntervalQuery(asset *pipeline.Asset, query string) ([]string, error) {
```
y-bruin, line 136: If this function is commonly used across all databases, let's create a shared utility function.

karakanb (approved):

PR #573, opened 2025-04-03
Interval modifiers

terzioglub:


pkg/pipeline/pipeline.go
```diff
@@ -600,32 +601,37 @@ func (s AthenaConfig) MarshalJSON() ([]byte, error) {
...
+	Secrets           []SecretMapping    `json:"secrets" yaml:"secrets,omitempty" mapstructure:"secrets"`
+	Extends           []string           `json:"extends" yaml:"extends,omitempty" mapstructure:"extends"`
+	Columns           []Column           `json:"columns" yaml:"columns,omitempty" mapstructure:"columns"`
+	CustomChecks      []CustomCheck      `json:"custom_checks" yaml:"custom_checks,omitempty" mapstructure:"custom_checks"`
+	Metadata          EmptyStringMap     `json:"metadata" yaml:"metadata,omitempty" mapstructure:"metadata"`
+	Snowflake         SnowflakeConfig    `json:"snowflake" yaml:"snowflake,omitempty" mapstructure:"snowflake"`
+	Athena            AthenaConfig       `json:"athena" yaml:"athena,omitempty" mapstructure:"athena"`
+	upstream          []*Asset
+	downstream        []*Asset
+	IntervalModifiers IntervalModifiers `json:"interval_modifiers" yaml:"interval_modifiers,omitempty" mapstructure:"interval_modifiers"`
```
karakanb, line 634: the exported fields have been grouped together at the top, then comes the unexported fields like `upstream` and `downstream`, I suggest grouping them together.

pkg/pipeline/pipeline.go
```diff
@@ -600,32 +601,37 @@ func (s AthenaConfig) MarshalJSON() ([]byte, error) {
 	return json.Marshal(s)
 }
 
+type IntervalModifiers struct {
+	Start string `json:"start" yaml:"start" mapstructure:"start"`
+	End   string `json:"end" yaml:"end" mapstructure:"end"`
```
karakanb, line 606: I think we should do the parsing logic once and convert these into a proper structure afterwards, meaning that the resulting struct should be something like this:

```go
type TimeModifier struct {
	Days int `json:"days"`
	Hours int `json:"hours"`
	Minutes int `json:"minutes"`
	Seconds int `json:"seconds"`
	CronPeriods int `json:"cron_periods"`
}

func (t *TimeModifier) UnmarshalYAML(value *yaml.Node) error {
	// convert string to time modifier
}

type IntervalModifiers struct {
	Start TimeModifier `json:"start"`
	End   TimeModifier `json:"end"`
}
```

This would allow us to avoid having to repeat this same logic everywhere in our other applications.

pkg/pipeline/pipeline.go
```diff
@@ -1688,3 +1694,36 @@ func (b *Builder) SetAssetColumnFromGlossary(asset *Asset, pathToPipeline string
...
 }
+
+func ModifyDate(t time.Time, modifier string) (time.Time, error) {
+	if modifier == "" {
+		return t, nil
+	}
+
+	if len(modifier) < 2 {
+		return t, errors.New("invalid interval format; must contain a number + suffix (h, m, s, or M)")
+	}
```
karakanb, line 1705: that's not true, the user can give `-3`, which means `go back 3 periods`.

karakanb (approved):

PR #575, opened 2025-04-04
Update context with intervalModifers before jinja rendering 

terzioglub:


pkg/jinja/jinja.go
```diff
@@ -81,9 +84,61 @@ func NewRendererWithYesterday(pipelineName, runID string) *Renderer {
 	return NewRendererWithStartEndDates(&startDate, &endDate, pipelineName, runID)
 }
 
-func (r *Renderer) Render(query string) (string, error) {
-	r.queryRenderLock.Lock()
+func UpdateContextWithIntervalModifiers(ctx *exec.Context, modifiers pipeline.IntervalModifiers) error {
```
karakanb, line 87: this seems like a hack to me. we'll have to have asset-specific jinja rendering anyway, since we want to be able to pass various other variables like `{{this}}` moving forward, I think this suits well. Instead of modifying existing keys, I'd suggest creating a new jinja context with those dates, specific to the asset being executed.

PR #607, opened 2025-04-23
chore: update dependencies

terzioglub:
1.Updated various Go module dependencies to their latest versions.
2.Downgraded the ch-go dependency to v0.64.0 to maintain compatibility with clickhouse-go.
3.Replaced usage of golang.org/x/exp/maps.Keys() with a local keysFromMap function because:
The maps package from golang.org/x/exp is no longer supported after the update.
In Go 1.21, maps.Keys() returns an iterator


pkg/connection/connection.go
```diff
@@ -108,6 +109,14 @@ type Manager struct {
 	mutex           sync.Mutex
 }
 
+func CollectSlice[T any](seq iter.Seq[T]) []T {
```
turtleDev, line 112: There's a builtin function for this:
https://pkg.go.dev/slices#Collect

PR #622, opened 2025-04-25
Policies v0

turtleDev:
## Summary

This PR extends Bruin's linting feature by adding support for user-defined **custom lint rules** through a `policy.yml` file.

- Adds `rulesets` for grouping and applying rules conditionally using **selectors** (`path`, `asset`, `pipeline`, `tag`).
- Built-in rules (`asset_name_is_lowercase`, `asset_name_is_schema_dot_table`, etc.) can be referenced directly without redefinition.
- If no selectors are specified, the ruleset applies to **all jobs**.
- Rulesets require **all** specified selectors to match for the rules to apply.
- Supports custom rule definitions using [expr](https://expr-lang.org/docs/language-definition)
- The `policy.yml` is located by searching for the nearest git repository in the subtree of the working directory.

## Example `policy.yml`

```yaml
define:
  - name: asset_has_owner
    description: every asset should have an owner
    criteria: asset.Owner != ""

rulesets:
  - name: ruleset_1
    selector:
      - path: .*/foo/.*
      - tag: critical
    rules:
      - asset_has_owner
      - asset_name_is_lowercase
      - asset_has_description
```

## Notes

- This feature is backwards compatible: existing built-in linting continues to work without `policy.yml`.
- possible edge case: the policy file is resovled by looking for a git repository. If no git repositories are found, bruin will not load any policy definitions, even if the file is present in the subtree.


terzioglub (approved):

pkg/lint/policy.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+	}
+	return true, nil
+}
+
+func addBoundaryAnchors(pattern string) string {
+	if !strings.HasPrefix(pattern, "^") {
+		pattern = "^" + pattern
+	}
+	if !strings.HasPrefix(pattern, "$") {
```
turtleDev, line 279: Bug

terzioglub (approved):

terzioglub (approved):

terzioglub (approved):

PR #632, opened 2025-04-29
 parse sql and add limit

terzioglub:
Parses sql with sqlGlot and adds limit to the query from the user given input . If it pails to parse if falls back to our existing strategy

### DIFF
--- a/cmd/fetch.go
+++ b/cmd/fetch.go
@@ -18,6 +18,7 @@ import (
 	"github.com/bruin-data/bruin/pkg/path"
 	"github.com/bruin-data/bruin/pkg/pipeline"
 	"github.com/bruin-data/bruin/pkg/query"
+	"github.com/bruin-data/bruin/pkg/sqlparser"
 	"github.com/bruin-data/bruin/pkg/telemetry"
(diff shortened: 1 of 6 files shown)


cmd/fetch.go
```diff
@@ -86,7 +87,18 @@ func Query() *cli.Command {
 				return handleError(c.String("output"), err)
 			}
 			if c.IsSet("limit") {
-				queryStr = addLimitToQuery(queryStr, c.Int64("limit"), conn)
+				parser, err := sqlparser.NewSQLParser(false)
```
albertobruin, line 90: initializing the parser is very expensive. Is there a way we can use a single parser through the whole application?

pkg/sqlparser/parser.go
```diff
@@ -308,3 +308,38 @@ func AssetTypeToDialect(assetType pipeline.AssetType) (string, error) {
 	}
 	return dialect, nil
 }
+
+func (s *SQLParser) AddLimit(sql string, limit int) (string, error) {
```
karakanb, line 312: let's add tests for this as well to ensure we are calling the right function in the right way from the Go side.

albertobruin (approved):

PR #646, opened 2025-05-11
EMR Serverless: Enable Quality Checks

turtleDev:
This PR adds support for running quality checks on EMR Serverless assets using AWS Athena. You can now define column and custom checks for your Spark jobs.

**How to enable:**
1. Create an Athena table over your output data.
2. Add an Athena connection in your `bruin.yml`.
3. Set `parameters.athena_connection` in your asset.

**Example asset:**
```yaml
name: users
type: emr_serverless.spark
columns:
  - name: id
    type: integer
    checks:
      - non_negative
custom_checks:
  - name: users are adults
    query: SELECT count(*) from users where age < 18
    value: 0
parameters:
  athena_connection: athena-default
```

### DIFF
--- a/docs/platforms/emr_serverless.md
+++ b/docs/platforms/emr_serverless.md
@@ -45,13 +45,13 @@ In order to stream logs, one of the following conditions must be met:
 > [!NOTE]
 > Python assets stream logs out-of-the-box. You don't need to specify `parameters.logs` for them.
 
-## Asset Types
+## EMR Serverless Assets
 
 Bruin supports two different ways of defining a Spark asset:
 - what we call a "managed" PySpark asset where Bruin takes care of delivering the code to the cluster as well
 - as an external asset defined with YAML where Bruin simply orchestrates
 
-### Python Asset
+### `emr_serverless.pyspark`
 A fully managed option where Bruin takes care of job setup, configuration, and execution. You only need to define the workload logic.
 
 * Supports only PySpark scripts.
@@ -150,7 +150,7 @@ Bruin uses this for:
 
 ![workspace diagram](media/pyspark-workspace.svg)
 
-### YAML Asset
+### `emr_serverless.spark`
 A lightweight option that only supports triggering a job. 
 
 * Supports both PySpark scripts and JARs.
--- a/docs/platforms/emr_serverless.md
+++ b/docs/platforms/emr_serverless.md
@@ -349,6 +349,18 @@ if __name__ == "__main__":
 ```
 :::
 
+::: tip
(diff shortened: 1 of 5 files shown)


terzioglub (dismissed):

docs/platforms/emr_serverless.md
```diff
@@ -179,6 +179,202 @@ This defines an asset that runs a spark job on an EMR Serverless Application def
...
+  - name: id          # [!code ++]
+    type: integer     # [!code ++]
+    checks:           # [!code ++]
+      - non_negative  # [!code ++]
+custom_checks: # [!code ++] 
+  - name: users are adults # [!code ++]
+    query: SELECT count(*) from users where age < 18 # [!code ++]
+    value: 0 # [!code ++]
+parameters: # [!code ++]
+  athena_connection: quality-tests # [!code ++]
```
karakanb, line 330: I suggest setting up the examples in such a way that they set this in the `pipeline.yml` by default.

docs/platforms/emr_serverless.md
```diff
@@ -204,6 +400,10 @@ parameters:
   # path where logs are stored or should be stored (optional)
   logs: s3://amzn-test-bucket/logs
 
+  # name of your athena connection (optional, defaults to "athena-default")
```
karakanb, line 403: if I am correct, this should default to whatever value set in `default_connections.athena` in pipeline.yml, which would then default to `athena-default`, is that right?

pkg/emr_serverless/checks.go
```diff
@@ -0,0 +1,100 @@
...
+
+type connectionRemapper struct {
+	connectionFetcher
+	ti scheduler.TaskInstance
+}
+
+func (cr *connectionRemapper) GetConnection(string) (interface{}, error) {
+	connectionName := cmp.Or(
+		cr.ti.GetAsset().Parameters["athena_connection"],
+		cr.ti.GetPipeline().DefaultConnections["athena"],
```
karakanb, line 89: I suggest calling `GetAthenaConnection` instead of this one, that way we do not replicate the connection finding logic here.

turtleDev: I wish we could. The connection resolution mechanism is implemented by `pipeline.GetConnectionNameForAsset` and not `connection.Manager`. Unless I'm missing something?

docs/platforms/emr_serverless.md
turtleDev: I felt that the example has too many pieces already, so I didn't want to add to that.

Instead, I've added a tip for using default values.

terzioglub (approved):

PR #662, opened 2025-05-16
chore/make rendering more maintainable

karakanb:
- **move CloneForAsset from extractor into renderer**
- **use renderer instead of extractor in custom checks**
- **chore: format**

### DIFF
--- a/pkg/ansisql/checks.go
+++ b/pkg/ansisql/checks.go
@@ -9,7 +9,6 @@ import (
 	"github.com/bruin-data/bruin/pkg/query"
 	"github.com/bruin-data/bruin/pkg/scheduler"
 	"github.com/pkg/errors"
-	"github.com/spf13/afero"
 )
 
 type connectionFetcher interface {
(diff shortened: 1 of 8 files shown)


terzioglub (approved):

pkg/ansisql/checks.go
```diff
@@ -187,28 +186,26 @@ func (c *NegativeCheck) Check(ctx context.Context, ti *scheduler.ColumnCheckInst
...
 
 func NewCustomCheck(conn connectionFetcher) *CustomCheck {
-	queryExtractor := query.WholeFileExtractor{
-		Fs:       afero.NewOsFs(),
-		Renderer: jinja.NewRendererWithYesterday("your-pipeline-name", "your-run-id"),
-	}
-	return &CustomCheck{conn: conn, extractor: &queryExtractor}
+	// TODO: this needs to use an actual renderer instead of the yesterday, since this `NewRendererWithYesterday` does not honor
+	// the parameters passed to the `bruin run` command.
+	return &CustomCheck{conn: conn, renderer: jinja.NewRendererWithYesterday("your-pipeline-name", "your-run-id")}
```
karakanb, line 196: @turtleDev seems like you used this `NewCustomCheck` constructor in the EMR Serverless checks, but this seems to have been always using a jinja renderer that's not really accurate. this constructor seems to be only used in tests, you might need to refactor your implementation to use the correct renderer coming from above.

PR #672, opened 2025-05-20
Ddl snowflake

jbrand-dsp:
Added ddl-materialization for snowflake as well as unit tests to test functionality.


pkg/snowflake/materialization.go
```diff
@@ -154,3 +155,32 @@ func buildTimeIntervalQuery(asset *pipeline.Asset, query string) (string, error)
 
 	return strings.Join(queries, ";\n") + ";", nil
 }
+
+func BuildDDLQuery(asset *pipeline.Asset, query string) (string, error) {
```
karakanb, line 159: uppercase functions are exposed to other packages in Go, in this case I don't think you want to expose this.
```suggestion
func buildDDLQuery(asset *pipeline.Asset, query string) (string, error) {
```

jbrand-dsp: Thanks for the tip! Changed now.

albertobruin (approved):

PR #724, opened 2025-06-03
patch: no color

terzioglub:


karakanb:
I think this is a losing game trying to catch all the places we might use colors. There's a better way to do this, the library we use supports `[NO_COLOR](https://no-color.org/)` already, so just setting that env variable if that flag is given should make this just work everywhere.

PR #738, opened 2025-06-05
hotfix: time_modifier formatting

terzioglub:


karakanb (approved):
approving, but you definitely need at least a unit test for this.

PR #748, opened 2025-06-08
feat: patch fill from db

terzioglub:


cmd/patch.go
```diff
@@ -44,6 +55,53 @@ func updateAssetDependencies(ctx context.Context, asset *pipeline.Asset, p *pipe
...
+func fillColumnsFromDB(pp *ppInfo, fs afero.Fs, environment string, fullRefresh bool) (string, error) {
+	// If not full-refresh and asset already has columns, skip
+	if !fullRefresh && len(pp.Asset.Columns) > 0 {
+		return fillStatusSkipped, nil
+	}
+	_, conn, err := getConnectionFromPipelineInfo(pp, environment)
+	if err != nil {
+		return fillStatusFailed, fmt.Errorf("failed to get connection for asset '%s': %w", pp.Asset.Name, err)
+	}
+	if !pp.Asset.IsSQLAsset() {
```
karakanb: why do this only for SQL assets? I might have a Python asset that I want the fill the columns of, shouldn't matter.

cmd/patch.go
```diff
@@ -44,6 +55,53 @@ func updateAssetDependencies(ctx context.Context, asset *pipeline.Asset, p *pipe
...
+	querier, ok := conn.(interface {
+		SelectWithSchema(ctx context.Context, q *query.Query) (*query.QueryResult, error)
+	})
+	if !ok {
+		return fillStatusFailed, fmt.Errorf("connection for asset '%s' does not support schema introspection", pp.Asset.Name)
+	}
+	tableName := pp.Asset.Name
+	queryStr := fmt.Sprintf("SELECT * FROM %s LIMIT 0", tableName)
+	q := &query.Query{Query: queryStr}
+	result, err := querier.SelectWithSchema(context.Background(), q)
```
karakanb, line 98: We have a `GetDatabaseSummary` function, currently defined only for Postgres and Redshift. We'll need to expand that function to other databases anyway in order to support [devenvs](https://bruin-data.github.io/bruin/getting-started/devenv.html#developer-environments). I suggest using that instead of running a select on the table. It would also allow us to fetch additional metadata from them that select might not be able to return.

cmd/patch.go
```diff
@@ -131,16 +189,34 @@ func Patch() *cli.Command {
...
+
+						updatedAssets := []string{}
+						skippedAssets := []string{}
+						failedAssets := map[string]error{}
 						processedAssets := 0
-						successfulAssets := 0
+
 						for _, asset := range foundPipeline.Assets {
+							pp := &ppInfo{Pipeline: foundPipeline, Asset: asset, Config: cm}
+							status, err := fillColumnsFromDB(pp, afero.NewOsFs(), "", false)
```
karakanb: I would make this a completely separate command instead of `fill-dependencies` one.

cmd/patch.go
```diff
@@ -177,10 +253,175 @@ func Patch() *cli.Command {
...
+								errorPrinter.Printf("- '%s': %s\n", assetName, err)
+							}
+						}
+
+						return nil
+					}
+				},
+			},
+			{
+				Name:      "fill-from-db",
```
karakanb:
```suggestion
				Name:      "fill-columns-from-db",
```

cmd/patch.go
terzioglub: its a seperate command they are both subcommands of bruin patch , should we seperate this one from the patch family ?

cmd/patch.go
terzioglub: what other metadata would be in the scope of this command ?  description comes to mind but it seems like getDatabase doesnt return the description, should we extend getDatabase to return Descriptions?

terzioglub: also `selectWithSchema `already returns the column type and all of the column names, which is useful

karakanb (approved):

PR #784, opened 2025-06-13
skip long running queries from lineage

terzioglub:


pkg/sqlparser/parser.go
```diff
@@ -138,6 +138,14 @@ type Lineage struct {
 }
 
 func (s *SQLParser) ColumnLineage(sql, dialect string, schema Schema) (*Lineage, error) {
+	if len(sql) > 10000 {
```
karakanb, line 141: you might want to make this configurable so that we can change that value in different contexts

PR #815, opened 2025-06-19
feat/quickbooks

sanjushahgupta:


pkg/quickbooks/config.go
```diff
@@ -0,0 +1,26 @@
...
+	ClientID     string `yaml:"client_id" json:"client_id" mapstructure:"client_id"`
+	ClientSecret string `yaml:"client_secret" json:"client_secret" mapstructure:"client_secret"`
+	RefreshToken string `yaml:"refresh_token" json:"refresh_token" mapstructure:"refresh_token"`
+	Environment  string `yaml:"environment,omitempty" json:"environment,omitempty" mapstructure:"environment"`
+	MinorVersion string `yaml:"minor_version,omitempty" json:"minor_version,omitempty" mapstructure:"minor_version"`
+}
+
+// GetIngestrURI builds an ingestr URI for QuickBooks.
+func (c *Config) GetIngestrURI() string {
+	uri := fmt.Sprintf("quickbooks://?company_id=%s&client_id=%s&client_secret=%s&refresh_token=%s",
```
turtleDev: Can we use the url package to build the URI? I just fixed a bug today that was due to an unescaped credential.

PR #820, opened 2025-06-21
feat/format with sqlfluff

karakanb:
This PR introduces a new `--sqlfluff` parameter into `bruin format` command that will automatically run SQLFluff on the SQL assets based on their dialects.

### DIFF
--- a/cmd/format.go
+++ b/cmd/format.go
@@ -188,6 +188,10 @@ func formatAsset(path string) (*pipeline.Asset, error) {
 		return nil, errors2.Wrap(err, "failed to build the asset")
 	}
 
+	if asset == nil {
+		return nil, errors2.New("no valid asset found in the file")
+	}
+
 	return asset, asset.Persist(afero.NewOsFs())
 }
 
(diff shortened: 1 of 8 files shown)


cmd/format.go
```diff
@@ -125,38 +136,48 @@ func Format(isDebug *bool) *cli.Command {
 				}
 				return cli.Exit("", 1)
 			}
-			if !checkLint && len(errorList) == 0 {
+
+			if checkLint || len(errorList) > 0 {
```
turtleDev, line 140: This will print `null` if lint check is requested but there are no errors.

cmd/format.go
```diff
@@ -222,3 +243,186 @@ func normalizeLineEndings(content []byte) []byte {
...
+
+	// Full pipeline - analyze all SQL files and detect dialects in parallel
+	sqlPaths := path.GetAllPossibleAssetPaths(repoPath, assetsDirectoryNames, []string{".sql"})
+
+	if len(sqlPaths) == 0 {
+		return []python.SQLFileInfo{}, nil
+	}
+
+	// Use conc pool to process assets in parallel with max 30 goroutines
+	assetPool := pool.New().WithMaxGoroutines(30)
```
turtleDev, line 345: Would be nice to have the concurrency configurable. Depending on the number of assets, a higher or lower number may offer better performance.

turtleDev (approved):

PR #823, opened 2025-06-23
feat: make SQL query length limit configurable

terzioglub:


pkg/sqlparser/parser.go
```diff
@@ -40,6 +41,10 @@ type SQLParser struct {
 }
 
 func NewSQLParser(randomize bool) (*SQLParser, error) {
```
turtleDev, line 43: We can use functional options. As it stands, the query limit is still hard-coded and cannot be overridden at runtime.

pkg/sqlparser/parser.go
```diff
@@ -24,12 +24,13 @@ import (
...
-	tmpDir      string
-	started     bool
-	randomize   bool
+	ep             *python.EmbeddedPython
+	sqlglotDir     *embed_util.EmbeddedFiles
+	rendererSrc    *embed_util.EmbeddedFiles
+	tmpDir         string
+	started        bool
+	randomize      bool
+	MaxQueryLength int
```
turtleDev, line 33: let's keep this unexported to keep it consistent with the rest of the fields?

PR #880, opened 2025-07-07
 add timeout flag

terzioglub:
cancel long running queries with timeout flag

### DIFF
--- a/cmd/fetch.go
+++ b/cmd/fetch.go
@@ -86,6 +86,12 @@ func Query() *cli.Command {
 				EnvVars: []string{"BRUIN_CONFIG_FILE"},
 				Usage:   "the path to the .bruin.yml file",
 			},
+			&cli.IntFlag{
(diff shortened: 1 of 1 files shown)


cmd/fetch.go
```diff
@@ -116,13 +122,28 @@ func Query() *cli.Command {
...
-				ctx := context.Background()
+				// Create context with timeout if specified
+				var ctx context.Context
+				var cancel context.CancelFunc
+				if timeout := c.Int("timeout"); timeout > 0 {
+					ctx, cancel = context.WithTimeout(context.Background(), time.Duration(timeout)*time.Second)
+					defer cancel()
+				} else {
+					ctx = context.Background()
+				}
```
karakanb, line 136: meh, this would work only if the CLI is still running. how do we attach this information to the underlying database queries as well, e.g. directly to the snowflake job?

PR #888, opened 2025-07-09
Add Environments Delete , Update 

mustafa3rsan:
# Update environment name and schema prefix
  bruin environments update --name dev --new-name development
  --schema-prefix "dev_"

  # Delete environment with confirmation
  bruin environments delete --name staging

  # Delete environment without confirmation
  bruin environments delete --name staging --force

  # JSON output support
  bruin environments update --name dev --new-name development
  --output json

  All commands integrate seamlessly with existing CLI patterns
   and include comprehensive error handling.

### DIFF
--- a/pkg/config/manager.go
+++ b/pkg/config/manager.go
@@ -1212,3 +1212,111 @@ func (c *Config) AddEnvironment(name, schemaPrefix string) error {
 
 	return nil
 }
+
+// UpdateEnvironment updates an existing environment in the configuration.
+// If the environment does not exist, an error is returned.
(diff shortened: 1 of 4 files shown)


pkg/config/manager.go
```diff
@@ -1212,3 +1212,111 @@ func (c *Config) AddEnvironment(name, schemaPrefix string) error {
 
 	return nil
 }
+
```
albertobruin, line 1215: Hi @mustafa3rsan , thanks for your contribution! Could we get some unit tests for this bit?

mustafa3rsan: okay i'll add

karakanb (approved):

PR #889, opened 2025-07-09
Add method to gather all secret names needed

albertobruin:


.vscode/launch.json
albertobruin, line 1: .vscode should be ignored and up to every engineer

PR #918, opened 2025-07-18
Mssql db summary

terzioglub:


pkg/mssql/config.go
```diff
@@ -47,3 +47,7 @@ func (c *Config) GetIngestrURI() string {
 
 	return u.String()
 }
+
+func (c *Config) GetDatabase() string {
```
albertobruin: What is the purpose of this method? You are already exporting `Database`?

albertobruin (approved):

PR #970, opened 2025-07-29
Trino

terzioglub:


pkg/trino/operator.go
```diff
@@ -0,0 +1,63 @@
...
+	for _, q := range splitQueries {
+		if strings.TrimSpace(q) == "" {
+			continue
+		}
+		query := &query.Query{Query: q}
+		err = conn.RunQueryWithoutResult(ctx, query)
+		if err != nil {
+			return err
+		}
+	}
```
karakanb: `queries` is there for this reason already, we have a slightly smarter query splitter that does all these for you, the operator should not care about this.

karakanb (approved):

PR #972, opened 2025-07-30
Improve workflow tests for bigquery cloud integration tests

jbrand-dsp:
Tests now execute from temp directory. For each workflow test, a temp directory is created. The test pipelines content is copied into this directory and then the test executes the workflow from there. This allows for parallel test execution.


integration-tests/cloud-integration-tests/bigquery/bigquery_test.go
```diff
@@ -70,195 +72,423 @@ func TestBigQueryWorkflows(t *testing.T) {
...
 
-			err := tt.workflow.Run()
-			require.NoError(t, err, "Workflow %s failed: %v", tt.workflow.Name, err)
+			tempDir, tempErr := os.MkdirTemp("", "bigquery-test-*")
+			require.NoError(t, tempErr, "Failed to create temporary directory")
+			defer os.RemoveAll(tempDir)
+
+			srcPipelines := filepath.Join(currentFolder, "test-pipelines")
+			destPipelines := filepath.Join(tempDir, "test-pipelines")
+			copyErr := copyDir(srcPipelines, destPipelines)
```
turtleDev: you can reuse the `err` variable. There's no need to create separate error variables, unless you need to reference errors later.

```go
tmpDir, err := os.MkdirTemp("", "bigquery-test-*")
if err != nil { ... }

err = copyDir(srcPipelines, destPipelines)
if err != nil { ... }
```

is perfectly idiomatic code.

PR #979, opened 2025-08-01
Added uv cache clean to bruin clean

jbrand-dsp:
Added new flag to clean uv-cache:

`bruin clean --uv-cache`

Prompts user with "Are you sure you want to clean uv cache? (y/N): "

### DIFF
--- a/cmd/clean.go
+++ b/cmd/clean.go
@@ -1,9 +1,13 @@
 package cmd
 
 import (
+	"bufio"
+	"fmt"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
+	"strings"
 
 	"github.com/bruin-data/bruin/pkg/git"
 	"github.com/bruin-data/bruin/pkg/telemetry"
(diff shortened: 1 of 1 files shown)


terzioglub (approved):

cmd/clean.go
```diff
@@ -41,7 +54,17 @@ type CleanCommand struct {
...
 }
 
-func (r *CleanCommand) Run(inputPath string) error {
+func (r *CleanCommand) Run(inputPath string, cleanUvCache bool) error {
+	// Clean uv caches if requested
+	if cleanUvCache {
+		if err := r.cleanUvCache(); err != nil {
+			return err
+		}
+		// Return early after uv cache cleaning, skip log cleanup
```
karakanb: why so? I think this is counter-intutive, instead I would expect this to clean bruin logs _and_ clean the uv cache.

cmd/clean.go
```diff
@@ -79,3 +102,66 @@ func (r *CleanCommand) Run(inputPath string) error {
...
+		// Prompt user to install uv if not found
+		if r.confirmUvInstall() {
+			// Use EnsureUvInstalled which will handle the installation
+			uvBinaryPath, err = uvChecker.EnsureUvInstalled(ctx)
+			if err != nil {
+				return errors.Wrap(err, "failed to install uv")
+			}
+		} else {
+			return errors.New("uv is not installed. Please install uv first using the bruin installation process")
+		}
```
karakanb: if uv doesn't exist simply inform the user and move on, there's no need to install uv just to be able to clean its cache.

PR #987, opened 2025-08-04
Add suggestion for command

mustafa3rsan:
add suggestions function for main commands and test

### DIFF
--- a/cmd/suggestions.go
+++ b/cmd/suggestions.go
@@ -3,6 +3,8 @@ package cmd
 import (
 	"fmt"
 	"os"
+
+	"github.com/urfave/cli/v2"
 )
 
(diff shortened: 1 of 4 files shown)


cmd/suggestions.go
```diff
@@ -0,0 +1,95 @@
...
+	return bestMatch
+}
+
+// HandleInvalidCommand prints a suggestion or generic error message
+func HandleInvalidCommand(invalidCmd string) {
+	validCommands := []string{
+		"validate", "run", "render", "lineage", "clean", "format",
+		"docs", "init", "environments", "query", "patch", "data-diff",
+		"diff", "import", "version", "help", "h", "connections",
+	}
```
karakanb: is there not a way to get these commands automatically?

PR #990, opened 2025-08-04
make domains as an array instead of map

batikankarakan:


cmd/internal.go
```diff
@@ -97,6 +100,66 @@ func ParsePipeline() *cli.Command {
 	}
 }
 
+func LoadGlossary() *cli.Command {
+	return &cli.Command{
+		Name:      "load-glossary",
```
karakanb: let's make this `parse-glossary`

karakanb (approved):

PR #997, opened 2025-08-05
allow flags to be anywhere

karakanb:
This PR bumps the urfave/cli version to v3, which has the very nice change of allowing flags to be used anywhere, so that this works now:

image


cmd/datadiff.go
```diff
@@ -157,7 +157,7 @@ func DataDiffCmd() *cli.Command {
 				return fmt.Errorf("failed to get connection '%s'", conn2Name)
 			}
 
-			ctx := c.Context
+			// ctx is already available from the function signature
```
albertobruin, line 160: this comment is kinda unnecessary

PR #1009, opened 2025-08-07
Refactor Integration Tests to use Go Test Framework

jbrand-dsp:


.github/workflows/build-test.yml
```diff
@@ -73,6 +73,23 @@ jobs:
...
+      - name: Build Bruin executable
+        run: |
+          mkdir -p bin
+          if [[ "${{ matrix.platform }}" == "windows-2022" ]]; then
+            go build -o bin/bruin.exe .
+          else
+            go build -o bin/bruin .
+          fi
+        shell: bash
+      - name: Verify binary exists
```
albertobruin: lets remove this step

Makefile
```diff
@@ -29,6 +29,9 @@ build-no-duckdb: deps
 	@echo "$(OK_COLOR)==> Building the application without DuckDB support...$(NO_COLOR)"
 	@CGO_ENABLED=0 go build -v -tags="bruin_no_duckdb" -ldflags="-s -w -X main.Version=$(or $(tag), dev-$(shell git describe --tags --abbrev=0)) -X main.telemetryKey=$(TELEMETRY_KEY)" -o "$(BUILD_DIR)/$(NAME)-no-duckdb" "$(BUILD_SRC)"
 
+# Default to serial execution
+ENABLE_PARALLEL ?=
```
albertobruin: we can also get rid of this line

Makefile
```diff
@@ -38,27 +41,53 @@ integration-test: build
...
-	@INCLUDE_INGESTR=1 go run integration-tests/integration-test.go
+	@cd integration-tests && go test -tags="no_duckdb_arrow" -v -count=1 $(if $(SERIAL),-parallel=1,) .
 
-integration-test-cloud: build
-	@touch integration-tests/cloud-integration-tests/.git
-	@touch integration-tests/cloud-integration-tests/bruin
-	@rm -rf integration-tests/cloud-integration-tests/.git
-	@rm integration-tests/cloud-integration-tests/bruin
-	@echo "$(OK_COLOR)==> Running cloud integration tests...$(NO_COLOR)"
+it-individual: integration-test-individual
```
albertobruin: lets get rid of this too

Makefile
```diff
@@ -38,27 +41,53 @@ integration-test: build
...
+	@touch integration-tests/.git
+	@touch integration-tests/bruin
+	@rm -rf integration-tests/.git
+	@rm integration-tests/bruin
+	@echo "$(OK_COLOR)==> Running integration tests...$(NO_COLOR)"
 	@cd integration-tests && git init
-	@cd integration-tests/cloud-integration-tests && go test -count=1 -v .
+	@cd integration-tests && go test -tags="no_duckdb_arrow" -v -count=1 -run ^TestIndividualTasks github.com/bruin-data/bruin/integration-tests $(if $(SERIAL),-parallel=1,)
+
+it-workflow: integration-test-workflow
```
albertobruin: and this

Makefile
```diff
@@ -38,27 +41,53 @@ integration-test: build
...
 	@mkdir -p integration-tests/duckdb-files  # Recreate the directory
 	@touch integration-tests/.git
 	@touch integration-tests/bruin
 	@rm -rf integration-tests/.git
 	@rm integration-tests/bruin
-	@echo "$(OK_COLOR)==> Running light integration tests...$(NO_COLOR)"
+	@echo "$(OK_COLOR)==> Running integration tests...$(NO_COLOR)"
 	@cd integration-tests && git init
-	@INCLUDE_INGESTR=0 go run integration-tests/integration-test.go
+	@cd integration-tests && INCLUDE_INGESTR=1 go test -tags="no_duckdb_arrow" -v -count=1 -run ^TestIngestrTasks github.com/bruin-data/bruin/integration-tests $(if $(SERIAL),-parallel=1,)
```
albertobruin: either use the home cooked parallel control or the native -parallel=1 but not both

Makefile
jbrand-dsp: deleted

Makefile
jbrand-dsp: deleted

Makefile
jbrand-dsp: deleted

Makefile
jbrand-dsp: Using homecooked, as t.parallel() is called when running individual tasks, workflow tasks and ingestr tasks (i.e. parallelises all three supersets of tests) as well as when running the tests themselves (i.e. individual tasks can run in parallel, workflow tasks can run in parallel etc). The flag enables parallel execution of everything. Using go flags would only override t.parallel() on the superset level.

.github/workflows/build-test.yml
jbrand-dsp: Windows e2e tasks fail without this build step. Verification step added for better clarity in case of error.

PR #1023, opened 2025-08-12
Fix athena-test-tempfiles

jbrand-dsp:
Fixed athena cloud integration tests to use temp directory to avoid file clean up after test run. All tests pass.


batikankarakan (approved):

albertobruin (commented):
i reviewed. It was great

PR #1033, opened 2025-08-15
Feat: table sensor for snowflake

jbrand-dsp:


pkg/lint/rules.go
```diff
@@ -1060,7 +1060,7 @@ func EnsureSnowflakeSensorHasQueryParameterForASingleAsset(ctx context.Context,
 
 func EnsureBigQueryTableSensorHasTableParameterForASingleAsset(ctx context.Context, p *pipeline.Pipeline, asset *pipeline.Asset) ([]*Issue, error) {
```
albertobruin: can we change the name of the function

pkg/ansisql/operator.go
```diff
@@ -15,6 +15,21 @@ import (
...
+type TableExistsChecker interface {
+	Select(ctx context.Context, q *query.Query) ([][]interface{}, error)
+	BuildTableExistsQuery(tableName string) (string, error)
+}
+
+type PipelineProvider interface {
+	GetConnectionNameForAsset(asset *pipeline.Asset) (string, error)
+}
+
+type TimeProvider interface {
```
albertobruin: where is this used?

pkg/lint/rules.go
jbrand-dsp: Was just thinking the same thing. I'm on it!

jbrand-dsp: Changed1

pkg/ansisql/operator.go
jbrand-dsp: Deleted

jbrand-dsp: I was meant to delete it in a previous commit where I was cleaning up the code

PR #1059, opened 2025-08-21
Feature: Postgres Table Sensor 

jbrand-dsp:
Implemented table sensor for Postgres.

Added unit testing for implementation. All pass.

Added integration tests. All pass.


pkg/postgres/db.go
```diff
@@ -771,3 +771,27 @@ func (c *Client) PushColumnDescriptions(ctx context.Context, asset *pipeline.Ass
 
 	return nil
 }
+
+func (c *Client) BuildTableExistsQuery(tableName string) (string, error) {
+	tableComponents := strings.Split(tableName, ".")
+	for _, component := range tableComponents {
+		if component == "" {
+			return "", fmt.Errorf("table name must be in format schema.table, '%s' given", tableName)
```
albertobruin: is this a product requirement?

pkg/postgres/db.go
```diff
@@ -771,3 +771,27 @@ func (c *Client) PushColumnDescriptions(ctx context.Context, asset *pipeline.Ass
...
+		}
+	}
+
+	if len(tableComponents) != 2 {
+		return "", fmt.Errorf("table name must be in format schema.table, '%s' given", tableName)
+	}
+
+	schemaName := strings.ToLower(tableComponents[0])
```
albertobruin: are you sure postgres is case insensitive? Also if it is, why are you lowercasing it anyway?

jbrand-dsp: Correct, it is not case sensitive. Deleted ToLower().

pkg/postgres/db.go
jbrand-dsp: Actually, no. If no schema is provided, Postgres assumes the schema to be "public". I've changed the code to reflect this logic.

jbrand-dsp: For reference, the parameter tableName is passed from the the table sensor asset's madatory parameter "table" (there aren't any constraints on this name)

albertobruin (approved):

PR #1065, opened 2025-08-21
Feature: Table Sensor Redshift

jbrand-dsp:
Implemented table sensor for Redshift.

Slight refactor/rewrapping of BuildTableExistsQuery() - since it is an interface method. 

Added unit testing for implementation. All pass.

Added integration tests. All pass.

### DIFF
--- a/pkg/postgres/db.go
+++ b/pkg/postgres/db.go
@@ -773,6 +773,14 @@ func (c *Client) PushColumnDescriptions(ctx context.Context, asset *pipeline.Ass
 }
 
 func (c *Client) BuildTableExistsQuery(tableName string) (string, error) {
+	// Check if this is a Redshift config by checking the type
+	if _, isRedshift := c.config.(*RedShiftConfig); isRedshift {
+		return c.BuildTableExistsQueryForRedshift(tableName)
+	}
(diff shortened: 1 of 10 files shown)


pkg/postgres/db.go
```diff
@@ -800,3 +808,32 @@ func (c *Client) BuildTableExistsQuery(tableName string) (string, error) {
 	)
 	return strings.TrimSpace(query), nil
 }
+
+func (c *Client) BuildTableExistsQueryForRedshift(tableName string) (string, error) {
```
karakanb, line 812: I suggest moving this into a new `redshift` package instead of doing it with an if-else here. 

We know that Redshift and Postgres are very similar, that's why we built them on the same principles, but we also foresaw that there would be enough differences that we would have to have separate asset types, such as `pg.sql` and `rs.sql`. I feel like this is one of those differences.

In the end, we don't have to rebuild/copy everything into the Redshift implementation, we could simply implement a new client/operator specific to redshift table sensor that would reuse whatever it needs from postgres, but represents itself to the rest of the codebase as an independent, Redshift-specific entity. This would allow us to keep these smaller pieces composable.

I think this could also be a good exercise for you.

PR #1068, opened 2025-08-22
Table Sensor for Redshift

jbrand-dsp:
Implemented table sensor for Redshift. Added unit and integration tests. All tests passing.

Explanation of implementation:

1. Entry Point (cmd/run.go)
- Line 1164: rsTableSensor := redshift.NewTableSensor(conn, sensorMode, wholeFileExtractor)
- Creates a Redshift table sensor using the connection manager, sensor mode, and file extractor
- This is the main entry point where Redshift table sensor functionality is initialized

2. Table Sensor Creation (pkg/redshift/operator.go)
- NewTableSensor() function creates a TableSensorConnectionWrapper
- Wrapper implements the connection interface that ansisql.NewTableSensor expects
- Returns a standard ansisql.TableSensor that can be used by the execution engine

3. Connection Wrapper (pkg/redshift/operator.go)
- TableSensorConnectionWrapper.GetConnection() intercepts connection requests
- When a postgres.Client is requested, it wraps it in a redshift.TableSensorClient
- This allows the generic ansisql table sensor to use Redshift-specific behavior


pkg/redshift/db.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/bruin-data/bruin/pkg/postgres"
+)
+
+type PostgresClientCreator interface {
```
albertobruin: I would rather not have references to `Postgres`  in the `redshift` package

albertobruin: Also what is the purpose of this interface?

pkg/redshift/operator.go
```diff
@@ -0,0 +1,33 @@
...
+	conn config.ConnectionGetter
+}
+
+func (w *TableSensorConnectionWrapper) GetConnection(name string) any {
+	conn := w.conn.GetConnection(name)
+	if conn == nil {
+		return nil
+	}
+
+	postgresClient, ok := conn.(*postgres.Client)
```
albertobruin, line 27: what is the point of casting the connection and then returning the uncasted connection if the casting fails? could this function not be just a one liner 
```
return w.conn.GetConnection(name)
```

jbrand-dsp: The casting logic exists because Redshift uses different system tables than PostgreSQL for checking table existence.
When it's a PostgreSQL client: Wrap it in TableSensorClient to use Redshift-specific SQL.

PR #1078, opened 2025-08-28
add docs for mcp

mustafa3rsan:


docs/getting-started/bruin-mcp.md
```diff
@@ -0,0 +1,69 @@
+# Bruin MCP (Model Context Protocol)
+
+Bruin MCP is a feature that acts as a bridge between AI editors (like Cursor) and the Bruin CLI. This feature allows AI assistants to directly access Bruin documentation and provide better assistance when creating pipelines.
```
terzioglub: I wouldn't call mcp a feature try something like :
_Bruin MCP is an integration layer that connects AI editors (like Cursor) with  Bruin CLI. It enables AI assistants to directly access Bruin documentation and provide better guidance when building pipelines_

docs/getting-started/bruin-mcp.md
```diff
@@ -0,0 +1,69 @@
...
+
+
+## Setup in Claude Code
+
+To use Bruin MCP in Claude Code, run the following command (works in bash, zsh, and other shells):
```
terzioglub: you don’t need to list out every shell unless there’s shell-specific behavior try 
_To use Bruin MCP in Claude Code, run the following command in your terminal:_

PR #1107, opened 2025-09-10
Add pipeline start date functionality

mustafa3rsan:


cmd/run.go
```diff
@@ -999,6 +1026,35 @@ func ParseDate(startDateStr, endDateStr string, logger logger.Logger) (time.Time
 	return startDate, endDate, nil
 }
 
+func DetermineStartEndDates(cliStartDate, cliEndDate string, pipeline *pipeline.Pipeline, fullRefresh bool, logger logger.Logger) (time.Time, time.Time, error) {
```
terzioglub: why not make this determinestart date only ? end date seems like a constant?

cmd/run.go
```diff
@@ -999,6 +1026,35 @@ func ParseDate(startDateStr, endDateStr string, logger logger.Logger) (time.Time
...
+	var startDate, endDate time.Time
+	var err error
+
+	// Start date logic
+	if fullRefresh && pipeline != nil && pipeline.StartDate != "" {
```
terzioglub: you can make this a bit more readable by  separating the and conditions e.g  if pipeline is nil early exit , if not full   refresh early exit etc but its a styling decision , I am leaving it up to you

cmd/run.go
```diff
@@ -455,6 +455,18 @@ var (
 	}
 )
 
+func ignoreApplyIntervalModifiers(c *cli.Command) (fullRefresh bool, applyIntervalModifiers bool) {
```
albertobruin: i think function should have a different name?

albertobruin: also as you know I'm not a fan of adding even more logic to run.go

mustafa3rsan: sabri and burak want to the function do you have any suggestions for use both full refresh and apply modifier?

PR #1121, opened 2025-09-16
Refactor: Table sensor validation rules

jbrand-dsp:


pkg/lint/rules.go
```diff
@@ -1070,32 +1070,102 @@ var TableSensorAllowedAssetTypes = map[pipeline.AssetType]bool{
...
+	tableItems := strings.Split(tableName, ".")
+	platformName := getPlatformNameFromAssetType(assetType)
+
+	for _, component := range tableItems {
+		if component == "" {
+			return fmt.Sprintf("%s table sensor `table` parameter contains empty components, '%s' given", platformName, tableName)
+		}
+	}
+
+	switch platformName {
```
turtleDev: A table-drive validator would be much easier to understand and maintain.

pkg/lint/rules.go
```diff
@@ -1070,32 +1070,102 @@ var TableSensorAllowedAssetTypes = map[pipeline.AssetType]bool{
...
-			Description: "BigQuery table sensor `table` parameter must be either in the format `dataset.table` or `project.dataset.table`",
+			Description: validationError,
 		})
 	}
 	return issues, nil
 }
 
-func EnsureBigQueryQuerySensorHasTableParameterForASingleAsset(ctx context.Context, p *pipeline.Pipeline, asset *pipeline.Asset) ([]*Issue, error) {
+func getPlatformNameFromAssetType(assetType pipeline.AssetType) string {
+	platformNames := map[pipeline.AssetType]string{
```
turtleDev: protip: local variables are allocated on every call. 

If you move it to package scope, it's only allocated once.

jbrand-dsp: Of course, good point!

pkg/lint/rules.go
jbrand-dsp: Suggestion integrated!

PR #1130, opened 2025-09-17
Fix: duckdb float

jbrand-dsp:
query command now converts the DuckDB tuple to a float value e.g. whereas before query would return `{3 2 299}` now it return `2.99`.

Added integration tests and adapted expectations of current integration tests.

### DIFF
--- a/cmd/fetch.go
+++ b/cmd/fetch.go
@@ -138,11 +138,14 @@ func Query() *cli.Command {
 				if err != nil {
 					return handleError(c.String("output"), errors.Wrap(err, "query execution failed"))
 				}
+				// Determine connection type once
+				connectionType := getConnectionType(conn)
+
 				// Output result based on format specified
 				inputPath := c.String("asset")
 				var resultsPath string
 				if c.Bool("export") {
-					resultsPath, err = exportResultsToCSV(result, inputPath)
+					resultsPath, err = exportResultsToCSV(result, inputPath, connectionType)
 					if err != nil {
 						return handleError(c.String("output"), errors.Wrap(err, "failed to export results to CSV"))
 					}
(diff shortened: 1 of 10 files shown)


cmd/fetch.go
```diff
@@ -124,6 +124,7 @@ func Query() *cli.Command {
 
 				queryStr = addLimitToQuery(queryStr, c.Int64("limit"), conn, parser, dialect)
 			}
+			//nolint:nestif
```
albertobruin, line 127: I think we should listen to lint here and unravel this if. just do the cast to interface and if not ok do 

```go
fmt.Printf("Connection type %s does not support querying.\n", c.String("connection"))
return nil
```

as you do below

albertobruin: always better to return early in the function and thus unravel the ifs and thens

cmd/fetch.go
```diff
@@ -138,11 +139,18 @@ func Query() *cli.Command {
 				if err != nil {
 					return handleError(c.String("output"), errors.Wrap(err, "query execution failed"))
 				}
+				var connectionType string
+				if assetType != "" {
```
albertobruin, line 143: is it possible to have assetType empty? I don't think so?

cmd/helpers.go
```diff
@@ -172,3 +175,227 @@ func printWarningForOutput(output string, message string) {
 		warningPrinter.Printf("%s\n", message)
 	}
 }
+
+// Asset type to connection type mapping.
+var assetTypeToConnectionMap = map[string]string{
```
albertobruin, line 180: I think we have something like this in pipeline.go

cmd/fetch.go
jbrand-dsp: It can indeed be empty in direct query mode. Wrote a test to prove so.

cmd/fetch.go
jbrand-dsp: I've unravelled this now

cmd/helpers.go
jbrand-dsp: Similar but not quite the same. I think it's slightly simpler to keep this map as it is.

PR #1139, opened 2025-09-22
Fix: Decimal printing for DuckDB and Postgres/Redshift

jbrand-dsp:
- Decimals are now printed in a readable format for `bruin query ` cmd output, e.g. `2.99` instead of `{3 2 299}` as it was before for DuckDB or` {299 -2 false finite true}` for Postgres/Redshift
- updated test expectations to fit implementation
- added new unit tests to test implementation

### DIFF
--- a/cmd/helpers.go
+++ b/cmd/helpers.go
@@ -186,7 +186,7 @@ func convertDuckDBDecimal(parts []string) (string, bool) {
 		return "", false
 	}
 
-	scale, err2 := strconv.ParseInt(parts[1], 10, 64)
+	scale, err2 := strconv.ParseInt(parts[1], 10, 0)
 	if err2 != nil {
 		return "", false
 	}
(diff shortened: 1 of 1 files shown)


cmd/helpers.go
```diff
@@ -172,3 +173,122 @@ func printWarningForOutput(output string, message string) {
...
+		return ""
+	}
+
+	// Check if this is a database decimal by looking at the string representation.
+	valStr := fmt.Sprintf("%v", val)
+	if strings.HasPrefix(valStr, "{") && strings.HasSuffix(valStr, "}") {
+		content := strings.Trim(valStr, "{}")
+		parts := strings.Fields(content)
+
+		if len(parts) == 3 && strings.Contains(connType, "duck.Client") {
```
terzioglub, line 258: looks like  the caller checks if the len is 3 and inside the function you make this check again one of them sees extra, same thing for postgres

cmd/helpers.go
```diff
@@ -172,3 +173,122 @@ func printWarningForOutput(output string, message string) {
...
+			if result, ok := convertDuckDBDecimal(parts); ok {
+				return result
+			}
+		}
+		if len(parts) == 5 && (strings.Contains(connType, "postgres.Client") || strings.Contains(connType, "redshift.Client")) {
```
terzioglub, line 263: can we potentially find a different way of detecting the correct type without using string. contains etc , if the underlying name changes  in the feature this will  stop working.  I think there might be some examples of this in the codebase . additionally a switch case pattern might be more go-native here

PR #1204, opened 2025-10-13
Add `PatchPipeline` Command to Internal CLI

DjamilaBaroudi:
# PR Overview 

This PR introduces the new **`PatchPipeline`** command to the internal CLI, enabling in-place updates of existing pipeline files.

### Changes

* Added `PatchPipeline` command for internal use.
* Added dedicated tests to verify patching behavior.


cmd/patch_pipeline_test.go
```diff
@@ -0,0 +1,296 @@
...
+	"github.com/bruin-data/bruin/pkg/path"
+	"github.com/bruin-data/bruin/pkg/pipeline"
+	errors2 "github.com/pkg/errors"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/urfave/cli/v3"
+)
+
+func createPatchPipelineCommand() *cli.Command {
```
terzioglub: why do we have the actual command  in the tests folder  ?

integration-tests/patch_pipeline_integration_test.go
jbrand-dsp, line 1: I don't quite understand these integration tests. Why are we testing retries? 

I would have a simple workflow test, if I understand the cmd correctly: We have a pipeline in the integration-tests/test-pipelines. We patch the pipeline and see if the pipeline that was created matches our expectations (depending on what we patch). We should create a directory for the patched pipeline in tempDir for this. Look at `"run_pipeline_with_scd2_by_column"` for creating this folder.

PR #1307, opened 2025-11-07
add update_on_merge functionality to duckdb

mustafa3rsan:


integration-tests/test-pipelines/duckdb-materialization-merge/assets/inventory.sql
```diff
@@ -10,10 +10,13 @@ columns:
     primary_key: true
   - name: item_name
     type: VARCHAR
+    update_on_merge: true
   - name: stock
     type: INTEGER
+    update_on_merge: true
   - name: unit_price
     type: DOUBLE
+    update_on_merge: true
```
terzioglub: dont add it to all columns so we can test a scenerio in which there is an update on an existing row  but since update_on_merge is not provided the row stays the same

pkg/duckdb/materialization.go
```diff
@@ -72,6 +72,19 @@ func buildIncrementalQuery(task *pipeline.Asset, query string) (string, error) {
 	return strings.Join(queries, ";\n") + ";", nil
 }
 
+func getColumnsWithMergeLogic(asset *pipeline.Asset) []pipeline.Column {
```
terzioglub: do we not have this helper already in pipeline.go or ansisql package ?

terzioglub (approved):

PR #1325, opened 2025-11-12
replace go-duckdb with the adbc driver

karakanb:
This PR replaces the cgo-based duckdb driver with the ADBC driver managed by dbc, speeding up the CLI in general, as well as simplifying our build process significantly.


pkg/python/uv.go
```diff
@@ -41,6 +41,7 @@ const (
 	pythonVersionForIngestr = "3.11"
 	ingestrVersion          = "0.14.101"
 	sqlfluffVersion         = "3.4.1"
+	adbcDriverManagerVersion = "1.2.0"
```
zeroshade: why are you installing this much older version of the driver manager? this should probably be 1.9.0 not 1.2.0

.github/workflows/build-test.yml
```diff
@@ -75,6 +75,10 @@ jobs:
       - name: Install winget
         if: matrix.platform == 'windows-2022'
         uses: Cyberboss/install-winget@v1
+      - name: Add MinGW to PATH on Windows
+        if: matrix.platform == 'windows-2022'
+        run: echo "/c/mingw64/bin" >> $GITHUB_PATH
```
zeroshade: From what I read on the issue, I don't believe this works (but I could be wrong)

.github/workflows/build-test.yml
```diff
@@ -75,6 +75,10 @@ jobs:
...
+        shell: bash
       - run: make integration-test
```
zeroshade: I'm pretty sure this needs to be: 

```yaml
- run: |
    [[ "${{runner.os}}" != 'Windows' ]] || export PATH="/c/mingw64/bin:$PATH"
    make integration-test
```

pkg/duckdb/ephemeral_db.go
```diff
@@ -5,58 +5,659 @@ package duck
...
+
+	return conn, db, nil
 }
 
-func (e *EphemeralConnection) QueryContext(ctx context.Context, query string, args ...any) (*sql.Rows, error) {
-	conn, err := sqlx.Open("duckdb", e.config.ToDBConnectionURI())
+// QueryContext implements the connection interface.
+//
+//nolint:ireturn // Returning interface type is by design for abstraction
+func (e *ADBCConnection) QueryContext(ctx context.Context, query string, args ...any) (Rows, error) {
```
zeroshade: Instead of going through all this effort, if you're wanting to use the database/sql interface instead of using the ADBC interfaces directly, we already have https://pkg.go.dev/github.com/apache/arrow-adbc/go/adbc@v1.9.0/sqldriver which wraps an ADBC driver and provides the database/sql interface for you. Though, obviously that's a row oriented interface so you're losing some of the benefits by not keeping the data in Arrow.

But at least you'd be able to get rid of a lot of this boilerplate

pkg/duckdb/ephemeral_db.go
```diff
@@ -5,58 +5,659 @@ package duck
...
+func (r *adbcRows) Next() bool {
+	if r.closed {
+		return false
+	}
+
+	r.currentRow++
+
+	if r.currentRecord == nil || r.currentRow >= r.currentRecord.NumRows() {
+		// Don't release the previous record yet - keep it in allRecords
+		// This prevents premature memory cleanup
```
zeroshade: Why avoid the early cleanup?

pkg/duckdb/ephemeral_db.go
```diff
@@ -5,58 +5,659 @@ package duck
...
+		return "BOOLEAN"
+	case "date32", "date64":
+		return "DATE"
+	case "timestamp":
+		return "TIMESTAMP"
+	default:
+		// For decimal types, uppercase and fix spacing: "decimal(5, 2)" -> "DECIMAL(5,2)"
+		if len(arrowType) >= 7 && arrowType[:7] == "decimal" {
+			// Replace "decimal" with "DECIMAL" and remove spaces after commas
+			result := "DECIMAL" + arrowType[7:]
```
zeroshade: For some systems, would this be `NUMBER`?

pkg/duckdb/ephemeral_db.go
```diff
@@ -5,58 +5,659 @@ package duck
...
+func (r *adbcRow) Scan(dest ...interface{}) error {
+	if r.err != nil {
+		return r.err
+	}
+
+	if len(dest) != len(r.values) {
+		return fmt.Errorf("expected %d destination arguments, got %d", len(r.values), len(dest))
+	}
+
+	for i, value := range r.values {
```
zeroshade: Is this logic being duplicated? Can it be factored out to a helper function? (Or eliminated entirely by using the sqldriver package?)

pkg/duckdb/ephemeral_db.go
```diff
@@ -5,58 +5,659 @@ package duck
...
+		return arr.Value(idx)
+	case *array.Float32:
+		return float64(arr.Value(idx))
+	case *array.Float64:
+		return arr.Value(idx)
+	case *array.String:
+		// Explicitly copy the string to ensure it's not a view into Arrow memory
+		s := arr.Value(idx)
+		// Force a copy by converting to []byte and back
+		return string(append([]byte(nil), []byte(s)...))
```
zeroshade: Why not just use `strings.Clone`?

pkg/duckdb/ephemeral_db.go
```diff
@@ -5,58 +5,659 @@ package duck
...
+	case *array.LargeString:
+		// Explicitly copy the string to ensure it's not a view into Arrow memory
+		s := arr.Value(idx)
+		// Force a copy by converting to []byte and back
+		return string(append([]byte(nil), []byte(s)...))
```
zeroshade: Use the `array.StringLike` interface to cover String, LargeString and StringView in one case

pkg/duckdb/ephemeral_db.go
```diff
@@ -5,58 +5,659 @@ package duck
...
+	case *array.Binary:
```
zeroshade: Use the `array.BinaryLike` interface to cover Binary/LargeBinary/BinaryView with one case

pkg/duckdb/ephemeral_db.go
```diff
@@ -5,58 +5,659 @@ package duck
...
+		return arr.Value(idx)
+	case *array.Time64:
+		return arr.Value(idx)
+	case *array.Decimal128:
+		val := arr.Value(idx)
+		dt := arr.DataType().(*arrow.Decimal128Type)
+		f := val.ToFloat64(dt.Scale)
+		// Format with exact precision to avoid floating-point representation issues
+		// For scale 2, format as "%.2f"
+		formatted := fmt.Sprintf("%.*f", dt.Scale, f)
```
zeroshade: Could you use the decimals ToString function for this? 

Should this be something we add to the actual decimal type?

zeroshade (commented):
The ideal would be to have the ADBC pieces factored out to a common reusable package and then reference it in the DuckDB package part so that it would be easy to expand beyond the DuckDB implementation in the future. But I also understand wanting to keep this isolated for now.

That said, I'm curious why you didn't use the existing sqldriver package in ADBC if you were going to force it back into the database/sql interface. (I link it below in my comments). If there's issues with that wrapper, I'd be happy to work with you and fix those issues so you can simplify your code. 

Otherwise there's just some comments and questions about how you're using the Arrow types. Your interactions with dbc and UV seem really cool

pkg/duckdb/ephemeral_db.go
karakanb: I didn't know about the sqldriver package, now trying that out, thanks!

PR #1336, opened 2025-11-14
tests , docs and mapping for elasticsearch

mustafa3rsan:


pkg/pipeline/pipeline.go
```diff
@@ -110,6 +110,7 @@ const (
 	AssetTypeRedash                 = AssetType("redash")
 	AssetTypeGoodData               = AssetType("gooddata")
 	AssetTypeS3KeySensor            = AssetType("s3.sensor.key_sensor")
+	AssetTypeElasticsearch          = AssetType("elasticsearch")
```
albertobruin, line 113: but we don't really have an elasticsearch asset type do we? We use it only for ingestr. Same with others like gooddata and so on

albertobruin (approved):

PR #1350, opened 2025-11-17
Materialization mapping for ingestr and python

mustafa3rsan:


pkg/lint/rules.go
```diff
@@ -487,6 +497,15 @@ func ValidatePythonAssetMaterialization(ctx context.Context, p *pipeline.Pipelin
 		})
 	}
 
+	if asset.Materialization.Strategy != "" && asset.Materialization.Strategy != pipeline.MaterializationStrategyNone {
```
albertobruin: how about 
```go
    if asset.Materialization.Strategy == "" || asset.Materialization.Strategy == pipeline.MaterializationStrategyNone {
		return issues, nil
	}
	
	if !python.IsPythonMaterializationStrategySupported(asset.Materialization.Strategy) {
		issues = append(issues, &Issue{
			Task:        asset,
			Description: fmt.Sprintf("Materialization strategy '%s' is not supported for Python assets. Supported strategies are: %s", asset.Materialization.Strategy, python.GetSupportedPythonStrategiesString()),
		})
	}
```

Always better to [return early](https://danp.net/posts/reducing-go-nesting/)  and also less nested blocks

PR #1358, opened 2025-11-18
added location and ADC to GCP JSON marshalling logic

jbrand-dsp:
This pull request makes a small update to the `GoogleCloudPlatformConnection` struct's JSON marshaling logic. It now includes the `location` and `use_application_default_credentials` fields in the marshaled JSON output, with the latter being converted to a string. Additionally, it imports the `strconv` package to support this conversion.

### DIFF
--- a/docs/ingestion/snapchat-ads.md
+++ b/docs/ingestion/snapchat-ads.md
@@ -75,6 +75,82 @@ These resources can fetch data for a specific ad account or all ad accounts in t
 | `creatives` | id | updated_at | merge | Retrieves all creatives for ad account(s). Supports `creatives:ad_account_id` |
 | `segments` | id | updated_at | merge | Retrieves all audience segments for ad account(s). Supports `segments:ad_account_id` |
 
+### Stats / Measurement Data
+
+Snapchat Ads source supports fetching stats/measurement data for campaigns, ad squads, ads, and ad accounts through dedicated stats resources.
+
+#### Stats Resources
+
+| Table | Inc Strategy | Details |
+|-------|--------------|---------|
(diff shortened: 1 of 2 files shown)


pkg/config/connections.go
```diff
@@ -69,6 +71,17 @@ func (c *GoogleCloudPlatformConnection) GetCredentials() *google.Credentials {
 }
 
 func (c GoogleCloudPlatformConnection) MarshalJSON() ([]byte, error) {
+	// If ADC is enabled, don't read or include service account fields as they're not used
+	if c.UseApplicationDefaultCredentials {
+		return json.Marshal(map[string]string{
+			"name":                                c.Name,
+			"project_id":                          c.ProjectID,
+			"location":                            c.Location,
+			"use_application_default_credentials": "true",
```
karakanb: I suggest always having the same keys available in the json and not having a difference in structure

pkg/config/connections.go
```diff
@@ -47,15 +47,17 @@ func (c GoogleCloudPlatformConnection) MarshalYAML() (interface{}, error) {
...
-		m["service_account_file"] = c.ServiceAccountFile
-	} else if c.ServiceAccountJSON != "" {
-		m["service_account_json"] = c.ServiceAccountJSON
+	} else {
+		// Include only one of ServiceAccountJSON or ServiceAccountFile, whichever is not empty
+		if c.ServiceAccountFile != "" {
+			m["service_account_file"] = c.ServiceAccountFile
+		} else if c.ServiceAccountJSON != "" {
+			m["service_account_json"] = c.ServiceAccountJSON
+		}
```
karakanb: no need for this, is there?

PR #1388, opened 2025-11-27
Introduce URI Dependency validation for a single asset 

DjamilaBaroudi:
# PR Overview 
This PR enhances the asset validation process by introducing uri dependency check. If an asset relies on external URIs, this new validation step ensures those dependencies are correctly accounted for.

### DIFF
--- a/pkg/lint/lint.go
+++ b/pkg/lint/lint.go
@@ -220,11 +220,8 @@ func (l *Linter) LintAsset(ctx context.Context, rootPath string, pipelineDefinit
 	// now the actual validation starts
 	for _, rule := range rules {
 		levels := rule.GetApplicableLevels()
-		if slices.Contains(levels, LevelCrossPipeline) {
+		if hasURIDeps && slices.Contains(levels, LevelCrossPipeline) {
 			// Cross-pipeline rules are only included when asset has URI dependencies
-			if !hasURIDeps {
-				continue
-			}
 			issues, err := rule.ValidateCrossPipeline(ctx, pipelines)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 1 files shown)


pkg/lint/lint.go
```diff
@@ -208,15 +208,42 @@ func (l *Linter) LintAsset(ctx context.Context, rootPath string, pipelineDefinit
...
+	}
 
-		if len(issues) > 0 {
-			pipelineResult.Issues[rule] = issues
+	// now the actual validation starts
+	for _, rule := range rules {
+		levels := rule.GetApplicableLevels()
+		if slices.Contains(levels, LevelCrossPipeline) {
+			// Cross-pipeline rules are only included when asset has URI dependencies
+			if !hasURIDeps {
```
turtleDev: You can combine this with the if statement above.
```go
if hasURIDeps && slices.Contains(levels, LevelCrossPipeline) {
```

This will eliminate unnecessary iterations of this loop.

pkg/lint/lint.go
```diff
@@ -208,15 +208,42 @@ func (l *Linter) LintAsset(ctx context.Context, rootPath string, pipelineDefinit
...
+			if err != nil {
+				return nil, err
+			}
+			// Only include issues for this asset
+			for _, issue := range issues {
+				if issue.Task == asset {
+					pipelineResult.Issues[rule] = append(pipelineResult.Issues[rule], issue)
+				}
+			}
+		} else if slices.Contains(levels, LevelAsset) {
```
turtleDev: you may want to remove the `else` here.

If a rule declares both cross pipeline and asset levels, only cross pipeline validation will be run against it.

turtleDev (approved):

PR #1424, opened 2025-12-10
add query for single check success and switch bold to fmt for error log

mustafa3rsan:
image AFTER
image BEFORE

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -222,9 +222,9 @@ func (t *AssetInstance) AddDownstream(task TaskInstance) {
 type ColumnCheckInstance struct {
 	*AssetInstance
 
-	parentID string
-	Column   *pipeline.Column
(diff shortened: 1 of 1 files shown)


cmd/run.go
```diff
@@ -1169,17 +1175,39 @@ func printErrorsInResults(errorsInTaskResults []*scheduler.TaskExecutionResult,
 }
 
 func printSingleCheckError(result *scheduler.TaskExecutionResult) {
-	infoPrinter.Println("\nCheck Failed")
-	infoPrinter.Println(strings.Repeat("-", 12))
+	fmt.Println("\nCheck Failed")
```
albertobruin, line 1178: can we use `errorPrinter` instead?

mustafa3rsan: In the agent we are running with --no-color, I think it is not correct

mustafa3rsan: error printer printing in red

PR #1433, opened 2025-12-16
added getIngestrURI() to databricks

jbrand-dsp:


albertobruin (approved):

pkg/databricks/config.go
```diff
@@ -35,3 +35,29 @@ func (c *Config) ToDBConnectionURI() string {
...
+	}
+	if c.Catalog != "" {
+		query.Add("catalog", c.Catalog)
+	}
+	if c.Schema != "" {
+		query.Add("schema", c.Schema)
+	}
+
+	u := &url.URL{
+		Scheme:   "databricks",
```
DjamilaBaroudi, line 56: Is this `Schema` or `Scheme`

jbrand-dsp: Good question. It is indeed `scheme` and not `schema` here. It refers to the platform.

PR #1438, opened 2025-12-18
Import command improvements and schemas

mustafa3rsan:


pkg/bigquery/db.go
```diff
@@ -1303,3 +1303,76 @@ func (d *Client) GetDatabaseSummary(ctx context.Context) (*ansisql.DBDatabase, e
...
+
+	mu := sync.Mutex{}
+	var errs []error
+
+	workers := max(runtime.NumCPU(), 8)
+	p := pool.New().WithMaxGoroutines(workers)
+
+	// Only iterate over requested schemas (datasets)
+	for _, schemaName := range schemas {
+		p.Go(func() {
```
terzioglub, line 1327: create a variable in the loop `schemaName := schemaName`  to avoid goroutines capturing the shared range variable. otherwise you will have multiple goroutines working on the same schema

terzioglub (approved):

PR #1449, opened 2025-12-29
Add RerunCooldown support to Asset and Pipeline with Airflow translation

DjamilaBaroudi:
# PR Overview 
This PR introduces a RerunCooldown field to both Asset and Pipeline structs and implements translation logic to generate the equivalent configuration for Airflow. It also updates comment parsing and YAML task definitions to support the new field, and adds integration tests with corresponding pipeline and asset configurations.

### Pipeline parse with rerun_cooldown
Screenshot 2025-12-29 at 10 34 46

### Asset parse with rerun_cooldown
Screenshot 2025-12-29 at 10 38 21

### DIFF
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -3167,7 +3167,7 @@ func TestMacros(t *testing.T) {
 					ExitCode: 0,
 					Contains: []string{
 						// Pipeline default rerun_cooldown
-						`"default":{"type":"","parameters":null,"secrets":null,"interval_modifiers":{"start":{"months":0,"days":0,"hours":0,"minutes":0,"seconds":0,"milliseconds":0,"nanoseconds":0,"cron_periods":0,"template":""},"end":{"months":0,"days":0,"hours":0,"minutes":0,"seconds":0,"milliseconds":0,"nanoseconds":0,"cron_periods":0,"template":""}},"rerun_cooldown":300}`, `"retries_delay":300`,
+						`"default":{"type":"","parameters":null,"secrets":null,"interval_modifiers":null,"rerun_cooldown":300}`, `"retries_delay":300`,
 						// Asset with explicit rerun_cooldown
 						`"name":"test_asset"`, `"rerun_cooldown":600`, `"retries_delay":600`,
 						// Asset that inherits from pipeline


pkg/pipeline/pipeline.go
```diff
@@ -2068,6 +2077,24 @@ func (b *Builder) MutatePipeline(ctx context.Context, pipeline *Pipeline) (*Pipe
 	return pipeline, nil
 }
 
+func (b *Builder) translatePipelineRetryConfig(ctx context.Context, pipeline *Pipeline) (*Pipeline, error) {
+	if pipeline == nil {
+		return pipeline, nil
```
terzioglub: return  nil, nil

pkg/pipeline/pipeline.go
```diff
@@ -1456,6 +1458,8 @@ type Pipeline struct {
 	Catchup            bool                   `json:"catchup" yaml:"catchup,omitempty" mapstructure:"catchup"`
 	MetadataPush       MetadataPush           `json:"metadata_push" yaml:"metadata_push,omitempty" mapstructure:"metadata_push"`
 	Retries            int                    `json:"retries" yaml:"retries,omitempty" mapstructure:"retries"`
+	RerunCooldown      *int                   `json:"rerun_cooldown,omitempty" yaml:"rerun_cooldown,omitempty" mapstructure:"rerun_cooldown"`
```
terzioglub: why did  we choose to have them as pointers ?

pkg/pipeline/pipeline.go
```diff
@@ -2116,6 +2143,27 @@ func (b *Builder) SetupDefaultsFromPipeline(ctx context.Context, asset *Asset, f
 	return asset, nil
 }
 
+func (b *Builder) translateRetryConfig(ctx context.Context, asset *Asset, foundPipeline *Pipeline) (*Asset, error) {
```
terzioglub, line 2146: does the asset level or pipeline level  cooldown take precedence ?  and normally when we have a configuration on asset level , say interval_modifiers we dont add a separate field in the pipeline struct for that but handle  in the defaults field of pipeline struct . Are we going on a different direction in this pr ?

terzioglub (approved):

PR #1470, opened 2026-01-05
Add Python cache ignores to `.gitignore` during bruin run

aryahassibi:
## Summary
Add python cache patterns to the project .gitignore during `bruin run` alongside existing log ignore updates.

### Justification
Running Python assets generates bytecode caches. This keeps new projects and their diffs are cleaner.

> [!NOTE]
> Minimal Downside.
> This hides Python bytecode if a project **intentionally** wants to commit it for packaging or deployment; they would need an explicit negate rule.
> I could not find any other downsides

| Before | Now |
|--------|-------|
|Screenshot 2026-01-05 at 14 37 54|Screenshot 2026-01-05 at 14 37 50|

## How to Test
```
path/to/bruin init default my-pipeline
path/to/bruin run ./bruin/my-pipeline
```
- generated `.gitignore` should have python should contain `__pycache__/` and `*.py[cod]` in addition to other file/directories. (only after `run`)

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -859,7 +859,7 @@ func Run(isDebug *bool) *cli.Command {
 					return cli.Exit("", 1)
 				}
 			}
-			
+
 			err = ensurePythonCacheGitignore(afero.NewOsFs(), repoRoot.Path)
 			if err != nil {
 				errorPrinter.Printf("Failed to add Python cache patterns to .gitignore: %v\n", err)


turtleDev:
You may need to change the pattern to to just `__pycache__`.

Using the wild card at the end makes it apply to top level directories only, it won't apply to deeply nested `__pycache__`'s

karakanb (dismissed):

PR #1494, opened 2026-01-08
databricks: oauth

karakanb:


docs/platforms/databricks.md
```diff
@@ -15,7 +19,7 @@ In order to work with Databricks you can add as a configuration item to `connect
           token: "your-databricks-token"
           path: "your-databricks-endpoint-path"
           host: "your-databricks-host"
-          port: "your-databricks-port"
+          port: 443
```
aryahassibi, line 22: wanted to ask if the port should be "hardcoded" in the md file. maybe a quick explanation that it is the `default` (seems like the default port value is 443)

pkg/databricks/config_test.go
```diff
@@ -35,3 +35,136 @@ func TestConfig_ToDSN(t *testing.T) {
...
+	// Verify OAuth M2M parameters are present
+	assert.Contains(t, result, "client_id=my-client-id")
+	assert.Contains(t, result, "client_secret=my-client-secret")
+	assert.Contains(t, result, "catalog=my_catalog")
+	assert.Contains(t, result, "schema=my_schema")
+	// Should NOT contain token:xxx@ format
+	assert.NotContains(t, result, "token:")
+	// Should have databricks scheme
+	assert.Contains(t, result, "databricks://")
+}
```
aryahassibi, line 170: was just wondering if it would be possible/better for the `GetIngestrURI` and `ToDBConnectionURI` tests to be in table format like the ones you wrote for `UseOAuthM2M`.

turtleDev (approved):

PR #1496, opened 2026-01-08
Feature/bruin enhance command

batikankarakan:
The bruin ai enhance command uses AI (Claude) to automatically improve Bruin asset definition files by adding:
- Descriptions for assets and columns
- Data quality checks (not_null, unique, accepted_values, etc.)
- Tags and metadata


pkg/enhance/enhancer.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/bruin-data/bruin/pkg/pipeline"
+	"github.com/pkg/errors"
+)
+
+const (
+	defaultModel = "claude-sonnet-4-20250514"
+)
+
+// Enhancer coordinates the AI enhancement process for assets.
+type Enhancer struct {
```
terzioglub, line 30: it could be a nice idea to define a enhancerInterface   and enhancer would implement that this would help you  create a MockEnhancer struct and use that for testing

pkg/enhance/enhancer.go
```diff
@@ -0,0 +1,235 @@
...
+	prompt := BuildEnhancePrompt(asset.DefinitionFile.Path, asset.Name, pipelineName, tableSummaryJSON)
+	systemPrompt := GetSystemPrompt(tableSummaryJSON != "")
+
+	// Call Claude CLI - Claude will edit the file directly
+	if err := e.callClaude(ctx, prompt, systemPrompt); err != nil {
+		return errors.Wrap(err, "failed to enhance asset")
+	}
+
+	// Run bruin format on the file
+	if err := e.runBruinFormat(ctx, asset.DefinitionFile.Path); err != nil {
```
terzioglub: it looks like we are actually running the cli commands to format and validate the asset ,  It would be more ideal if we actually called the functions that they are using underneath , for instance after the ai finishes it job you can create assetFrom file again and call asset.Persist to format it correctly  and call the lint function on the asset directly to validate , this way we dont create subprocesses

pkg/enhance/enhancer.go
```diff
@@ -0,0 +1,235 @@
...
+
+	args = append(args, prompt)
+
+	cmd := exec.CommandContext(ctx, e.claudePath, args...) //nolint:gosec
+
+	if e.apiKey != "" {
+		cmd.Env = append(os.Environ(), "ANTHROPIC_API_KEY="+e.apiKey)
+	}
+
+	if e.debug {
```
terzioglub, line 142: is there a reason we need streaming  ?, it looks like  we  are waiting for the end  output anyways , if streaming is not actually needed you can just read the  stdout at the end and you could get rid of the piping etc logic

PR #1502, opened 2026-01-09
create a bigquery template

karakanb:


turtleDev (dismissed):

terzioglub (dismissed):

templates/bigquery/assets/example.sql
```diff
@@ -0,0 +1,59 @@
+/* @bruin
+
+materialization:
+  type: table
+
+columns:
+  - name: id
+    type: int64
```
arsalann: if i remember correctly, `int64` is not correct and it should be `integer` instead

templates/bigquery/assets/example.sql
```diff
@@ -0,0 +1,59 @@
...
+  - name: country
+    type: string
+    description: the country
+    primary_key: true
+    checks:
+      - name: not_null
+  - name: name
+    type: string
+    description: Just a name
+    update_on_merge: true
```
arsalann: is this necessary if there's no merge strategy? I don't think it would break but it might be confusing

templates/bigquery/assets/example.sql
```diff
@@ -0,0 +1,59 @@
+/* @bruin
+
+materialization:
+  type: table
```
arsalann, line 4: would be nice to also demonstrate cluster and partition

templates/bigquery/assets/example.sql
```diff
@@ -0,0 +1,59 @@
...
+    checks:
+      - name: not_null
+      - name: positive
+      - name: non_negative
+  - name: country
+    type: string
+    description: the country
+    primary_key: true
+    checks:
+      - name: not_null
```
arsalann, line 23: already demonstrated not_null check above, would be nice to demonstrate `nullable` tag

templates/bigquery/assets/macro_example.sql
```diff
@@ -0,0 +1,52 @@
...
+  - example
+
+columns:
+  - name: country
+    type: string
+    description: Country name
+  - name: count
+    type: int64
+    description: Number of people per country
+
```
arsalann, line 20: should we demonstrate a custom check here?

templates/bigquery/macros/aggregations.sql
```diff
@@ -0,0 +1,26 @@
...
+GROUP BY {{ column }}
+ORDER BY {{ order_by }} DESC
+{%- endmacro %}
+
+{% macro sum_by(table, group_column, sum_column) -%}
+SELECT
+    {{ group_column }},
+    SUM({{ sum_column }}) as total
+FROM {{ table }}
+GROUP BY {{ group_column }}
```
arsalann: can also just be `group by all` 😅 sorry just being picky

arsalann (approved):

PR #1697, opened 2026-02-18
Remove airport_fee from trips_summary.sql

mushroomsandchai:
Removed since we are not using the column anywhere in staging output, green taxi-trip dataset does not have an airport_fee column, and since duckdb is throwing a SQL error.


arsalann:
@mushroomsandchai this looks fine to me but I'm curious what SQL error were you getting? I have created and ran this pipeline many times and haven't experienced any errors yet

PR #1738, opened 2026-02-25
Hooks: lint/validate + format for SQL queries

aryahassibi:
## Summary 

- Added hook query dry-run validation for SQL assets in `bruin validate`.
- Each non-empty `hooks.pre` / `hooks.post` query is checked with the asset connection.
- `bruin format` normalizes hook SQL (`trim + trailing ;`).
- Validation also normalizes hook queries to the same style before checking.
- Only BigQuery, Snowflake, and Postgres/Redshift clients support validation for now. (have the `IsValid()` method)

### `bruin validate` Now
Screenshot 2026-02-25 at 13 05 04

### Before
Screenshot 2026-02-25 at 13 07 22


pkg/lint/rules.go
```diff
@@ -1377,6 +1377,77 @@ func ValidateCustomCheckQueryDryRun(connections connectionManager, renderer jinj
...
+
+		for i, hook := range asset.Hooks.Post {
+			validateHook("post", i+1, hook.Query)
+		}
+
+		return issues, nil
+	}
+}
+
+func normalizeHookQueryForValidation(queryText string) string {
```
jbrand-dsp: This won't catch examples such as.

`SELECT 1;  -- This example comment comes after the semicolon`

The problem of removing semicolons has been often discussed in CLI. There is no simple solution to it, despite the problem appearing simple on the surface. We just advise users to avoid using the semicolon in this way.

cmd/testdata/unformatted-pipeline/assets/hook_needs_reformatting.sql
```diff
@@ -0,0 +1,16 @@
+/* @bruin
+
+name: format_test.hook_asset
+type: duckdb.sql
+
+hooks:
```
albertobruin: why does this need reformatting?

aryahassibi: quotation + space trimming (they already also happen by default during formatting)
plus, we add trailing `;` for queries

pkg/lint/rules.go
```diff
@@ -1377,6 +1377,77 @@ func ValidateCustomCheckQueryDryRun(connections connectionManager, renderer jinj
...
+	trimmed := strings.TrimSpace(queryText)
+	if trimmed == "" {
+		return ""
+	}
+	if strings.HasSuffix(trimmed, ";") {
+		return trimmed
+	}
+	return trimmed + ";"
+}
```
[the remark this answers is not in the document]
aryahassibi: formatStatement is not exported tho

cmd/testdata/unformatted-pipeline/assets/hook_needs_reformatting.sql
aryahassibi: or are you specifically asking about lines 3-6?

albertobruin (approved):

pkg/lint/rules.go
aryahassibi: yep makes sense
pushed a fix for this

PR #1836, opened 2026-03-24
Zoomcamp readme error

sotoblanco:
This should be cd bruin which matches with the folder you show in the video and documentation. cd my-first-pipeline cause error message as you need to go inside bruin folder first.

### DIFF
--- a/templates/zoomcamp/README.md
+++ b/templates/zoomcamp/README.md
@@ -157,7 +157,7 @@ Let's start by initializing a simple project to learn the basics before diving i
 **Initialize the default template:**
 ```bash
 bruin init default my-first-pipeline
-cd my-first-pipeline
+cd bruin
 ```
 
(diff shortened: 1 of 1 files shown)


templates/zoomcamp/README.md
```diff
@@ -157,7 +157,7 @@ Let's start by initializing a simple project to learn the basics before diving i
 **Initialize the default template:**
 ```bash
 bruin init default my-first-pipeline
-cd my-first-pipeline
+cd bruin
```
arsalann, line 160: this is correct only when the folder is not already git initialized
when the folder is not git initialized, it will create `bruin/my-first-pipeline`
when the folder is already git initialized, it will just create `my-first-pipeline`

I will approve this since that's the case most of the time for students - I will separately add a note there to clarify this.

arsalann (approved):

PR #1999, opened 2026-05-04
feat(lakehouse): support S3-compatible storage (endpoint, url_style, use_ssl)

### LINKED ISSUE #1994: Support S3-compatible storage (MinIO, Tigris, R2, B2) for DuckLake — endpoint, url_style, use_ssl on storage.s3 auth block
## Problem

Bruin's DuckDB lakehouse config currently exposes `storage.type: s3` with `region` + `auth.access_key`/`secret_key`/`session_token`, which works for AWS S3 but doesn't expose the fields needed for S3-compatible storage backends:

- `endpoint` (e.g. `minio.local:9000`, `https://<account>.r2.cloudflarestorage.com`, `https://fly.storage.tigris.dev`)
- `url_style` (path vs vhost — MinIO and several others require `path`)
- `use_ssl` (for plain-HTTP local dev)

DuckDB's underlying `httpfs` extension supports all three via `CREATE SECRET`, so this is a config surface gap in Bruin rather than a capability gap downstream.

## Use cases

- **MinIO** — local dev parity with cloud-object-store production (very common pattern in the DuckLake community; existing walkthroughs from Daniël du Toit, Data Lab Tech TV, etc.)
- **Cloudflare R2** — cost-optimised production storage (no egress fees)
- **Backblaze B2** — same
- **Tigris** — multi-region S3-compatible (Fly.io ecosystem)
- **On-prem / air-gapped customer deployments** — relevant in regulated industries (financial services, healthcare) where customer infra is required

## Proposed config

Extend the existing `storage.s3` block with three optional fields:

```yaml
storage:
  type: s3
  path: "s3://my-bucket/ducklake"
  region: "us-east-1"          # existing
  endpoint: "minio.local:9000" # NEW — optional
  url_style: "path"            # NEW — optional, default "vhost"
  use_ssl: false               # NEW — optional, default true
  auth:
    access_key: "${ACCESS_KEY}"
    secret_key: "${SECRET_KEY}"
```

When `endpoint` is unset, behaviour is unchanged (AWS S3). When set, Bruin passes it through to the DuckDB `CREATE SECRET` call.

## Equivalent in raw DuckDB

```sql
CREATE SECRET (
  TYPE S3,
  KEY_ID '...',
  SECRET '...',
  ENDPOINT 'minio.local:9000',
  USE_SSL false,
  URL_STYLE 'path'
);
```

## Why this matters

S3-compatible storage is dominant in the DuckLake adoption pattern (most community examples use MinIO for dev or R2/B2 for cost). Without these fields exposed in `.bruin.yml`, users have to either:

- Pre-create the DuckDB secret outside Bruin (breaks the "single source of truth" config model), or
- Patch Bruin locally

Adding three optional pass-through fields is a small change that unlocks a large slice of the DuckLake ecosystem.

## Happy to contribute

If the maintainers are open to it, I can put up a PR.

mgozdev:
## Summary

Closes #1994.

Adds three optional pass-through fields to the DuckDB lakehouse `storage.s3` block — `endpoint`, `url_style`, `use_ssl` — wiring them straight through to DuckDB's `CREATE SECRET`. When unset, behaviour is unchanged (AWS S3 defaults), so this is fully backwards-compatible.

```yaml
storage:
  type: s3
  path: "s3://ducklake/warehouse"
  endpoint: "minio.local:9000"   # NEW
  url_style: "path"              # NEW
  use_ssl: false                 # NEW
  auth:
    access_key: "${MINIO_ACCESS_KEY}"
    secret_key: "${MINIO_SECRET_KEY}"
```

This unblocks DuckLake on MinIO (local dev), Cloudflare R2, Backblaze B2, Tigris, and on-prem / air-gapped deployments — the dominant adoption pattern in the DuckLake community today.

## Changes

- `pkg/config/lakehouse.go` — three new optional fields on `StorageConfig` (`Endpoint`, `URLStyle`, `UseSSL *bool`); `IsZero()` updated.
- `pkg/duckdb/lakehouse.go` — `generateS3Secret` appends `ENDPOINT` / `URL_STYLE` / `USE_SSL` lines when set.
- `pkg/config/lakehouse_test.go` — new `TestStorageConfig_IsZero` covering all eight zero/non-zero combinations.
- `pkg/duckdb/lakehouse_test.go` — three new `generateS3Secret` cases: endpoint-only, full MinIO, Cloudflare R2.
- `docs/platforms/duckdb.md` — new "S3-compatible storage" subsection with field table + MinIO and R2 examples.

`UseSSL` is `*bool` so "unset" is distinguishable from explicit `false` — DuckDB's default is `true`, and we only emit the field when the user opts in either way.

## Test plan

- [x] `go test -tags=no_duckdb_arrow ./pkg/config/... ./pkg/duckdb/...` passes locally
- [x] `go vet -tags=no_duckdb_arrow ./pkg/config/... ./pkg/duckdb/...` clean
- [x] `gofmt -l` clean on the five touched files
- [ ] CI passes
- [ ] Manual smoke test: DuckLake against MinIO with `endpoint` + `url_style: path` + `use_ssl: false` (intend to run locally against the demo environment that motivated the issue)

### DIFF
--- a/docs/platforms/duckdb.md
+++ b/docs/platforms/duckdb.md
@@ -288,6 +288,41 @@ storage:
     session_token: "${AWS_SESSION_TOKEN}" # optional
 ```
 
+##### S3-compatible storage (MinIO, R2, B2, Tigris, on-prem)
+
+For S3-compatible backends, three optional fields can be set on the `storage` block. They are passed through to DuckDB's `CREATE SECRET` and behaviour is unchanged when they are omitted (AWS S3 defaults).
+
+| Field | Type | Required | Default | Description |
+|-------|------|----------|---------|-------------|
+| `endpoint` | string | No | (AWS S3) | Custom S3 endpoint, e.g. `minio.local:9000`, `<account>.r2.cloudflarestorage.com`, `fly.storage.tigris.dev` |
+| `url_style` | string | No | `vhost` | `path` is required for MinIO and several other S3-compatible backends |
+| `use_ssl` | bool | No | `true` | Set to `false` for plain-HTTP local dev (e.g. MinIO without TLS) |
+
+```yaml
+# MinIO local dev
+storage:
+  type: s3
+  path: "s3://ducklake/warehouse"
+  endpoint: "minio.local:9000"
+  url_style: "path"
+  use_ssl: false
+  auth:
+    access_key: "${MINIO_ACCESS_KEY}"
+    secret_key: "${MINIO_SECRET_KEY}"
+```
+
+```yaml
+# Cloudflare R2
+storage:
+  type: s3
+  path: "s3://my-bucket/ducklake"
+  region: "auto"
+  endpoint: "<account>.r2.cloudflarestorage.com"
+  auth:
+    access_key: "${R2_ACCESS_KEY_ID}"
+    secret_key: "${R2_SECRET_ACCESS_KEY}"
+```
+
 #### GCS
 
 Bruin currently supports explicit GCS HMAC credentials in the `auth` block (`access_key` and `secret_key`).
--- a/docs/platforms/duckdb.md
+++ b/docs/platforms/duckdb.md
@@ -318,6 +318,7 @@ storage:
   path: "s3://my-bucket/ducklake"
   region: "auto"
   endpoint: "<account>.r2.cloudflarestorage.com"
+  url_style: "path"
   auth:
     access_key: "${R2_ACCESS_KEY_ID}"
     secret_key: "${R2_SECRET_ACCESS_KEY}"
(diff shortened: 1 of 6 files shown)


mgozdev:
Thanks @greptile-apps — both addressed in 6a1931e:

**P1 — R2 example:** added `url_style: "path"`. R2 does technically accept vhost (Cloudflare advertises both styles), but path-style is the dominant recipe in DuckDB+R2 community walkthroughs and matches what most users will be copying from elsewhere. Consistency with the MinIO example is a nice bonus.

**P2 — `url_style` validation:** added a check against `{path, vhost}` in `LakehouseConfig.Validate()`, with three new test cases (path valid, vhost valid, unknown value rejected). Same fail-fast pattern as the existing format / catalog / storage type checks — invalid values now surface at config load instead of as a DuckDB `CREATE SECRET` error at query time.

karakanb (dismissed):

karakanb (approved):

PR #2001, opened 2026-05-05
feat(gong): verify SHA256 checksum of downloaded binary before installation

MusaMisto:
## What
Adds SHA256 checksum verification to the gong helper binary download flow.

Before: binary was downloaded and installed with no integrity check.
After:  download binary → download checksums.txt → verify SHA256 → install only if valid.
On any failure the temp file is discarded and the final path is never written.

## Changes
- `pkg/gong/checksum.go` — three stdlib-only helpers: `parseChecksumManifest`,
  `verifySHA256`, `isValidSHA256`. No new dependencies — crypto/sha256 is already
  in the Go standard library.

- `pkg/gong/gong.go` — wires verification into `downloadGong` between
  `tmpFile.Close()` and `os.Chmod`/`os.Rename`. This placement is deliberate:
  the file must be fully written before its hash can be read, and chmod/rename
  must never run on an unverified file. The existing `defer os.Remove(tmpPath)`
  handles cleanup on any failure path automatically — the final path is never
  created on a bad checksum.
  Adds `baseURL`/`installDir` as private zero-value fields rather than constructor
  parameters so the public API is completely unchanged; existing callers using
  `&Checker{}` see no behavior difference.

- `.goreleaser.yaml` — adds `checksum: name_template: "checksums.txt" /
  algorithm: sha256`. The filename must match the hardcoded path in
  `buildChecksumURL` (`/releases/{version}/checksums.txt`) — making it explicit
  in the release config prevents any future divergence.

- `SECURITY.md` — documents the new Binary Integrity behavior for users and
  maintainers. Supply-chain integrity policy belongs in SECURITY.md, not README
  or docs/, so it is independently linkable and not buried in installation steps.

## Tests
- 20 unit tests covering all helper functions and every error path (parse, verify,
  edge cases including uppercase hash, missing file, invalid expected value)
- 4 integration-style tests using `httptest.Server`: valid checksum installs,
  wrong checksum blocks, missing entry blocks, server failure blocks.
  The "binary does not exist at final path" assertion in the failure cases is the
  key safety proof — it confirms `os.Rename` never ran.

## Testing
- `go test ./pkg/gong/...`: passed (32/32)
- Full `pkg/...` suite: passed for all packages unrelated to pre-existing CGO
  build failures that reproduce identically on main (no C compiler on this
  Windows dev machine; race detector requires GCC)


pkg/gong/checksum.go
```diff
@@ -0,0 +1,109 @@
...
+// The expected file format is one entry per line:
+
+//
+
+//	<sha256hex>  <filename>
+
+//
+
+// Empty lines and lines starting with '#' are ignored.
+
```
karakanb: can you please clean up the extra lines around all of these?

pkg/gong/checksum.go
```diff
@@ -0,0 +1,109 @@
...
+		if filename != artifactName {
+			continue
+		}
+
+		if !isValidSHA256(checksum) {
+			return "", fmt.Errorf("invalid checksum for %s", artifactName)
+		}
+
+		return checksum, nil
```
karakanb: same here, please clean up extra spaces

karakanb (changes requested):
can you please fix all the extra spaces in the code?

pkg/gong/gong.go
```diff
@@ -134,65 +198,142 @@ func (g *Checker) ensureInstalled(ctx context.Context, version string) (string,
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed to download checksum file: server returned status %d", resp.StatusCode)
+	}
+
+	data, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read checksum file: %w", err)
+	}
```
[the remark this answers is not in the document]
MusaMisto: The issue has been addressed.

pkg/gong/gong.go
```diff
@@ -134,65 +198,142 @@ func (g *Checker) ensureInstalled(ctx context.Context, version string) (string,
...
 	_, _ = fmt.Fprintf(output, "Installing gong %s...\n", version)
+
 	_, _ = fmt.Fprintf(output, "This is a one-time operation.\n")
+
 	_, _ = fmt.Fprintf(output, "\n")
 
-	downloadURL := buildDownloadURL(version)
+	base := g.effectiveBaseURL()
+
+	artifactName := "gong_" + getArchName()
```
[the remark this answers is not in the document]
MusaMisto: The issue has been addressed.

pkg/gong/checksum.go
MusaMisto: Clean up around the code has been done as requested.

pkg/gong/checksum.go
MusaMisto: Clean up around the code has been done as requested.

PR #2092, opened 2026-05-25
add lakehouse support for ingestr

TanayBensuYurtturk:


turtleDev:
how would a user define a ducklake connection?

TanayBensuYurtturk:
@turtleDev Ducklake connection is already supported in the asset [https://getbruin.com/docs/bruin/platforms/duckdb.html#ducklake](https://getbruin.com/docs/bruin/platforms/duckdb.html#ducklake)

turtleDev (approved):

PR #2157, opened 2026-06-10
Merge catchup and catchup_mode into a single field

albertobruin:
## Summary

- Replaces `Pipeline.Catchup` (bool) + `Pipeline.CatchupMode` (string) with a single `Catchup` field backed by a new `CatchupMode` string type.
- The unmarshalers accept either a bool or a string on both YAML and JSON: `false`/missing/unknown → `""`, `true`/`"active"` → `"active"`, `"all"` → `"all"`.
- The value always serializes as a string (`""`, `"active"`, or `"all"`) on both encoders.
- Removes the now-unnecessary `validateCatchupMode` helper; invalid strings silently collapse to none instead of erroring.
- Updates the unit-test, persist, and integration-test expectation fixtures to the new shape.
- Expands the `Catchup` section in `docs/pipelines/definition.md` to document the new bool-or-string semantics.

## Test plan

- [x] `make format`
- [x] `make test`
- [ ] `make integration-test-light`
- [ ] CI green

### DIFF
--- a/docs/pipelines/definition.md
+++ b/docs/pipelines/definition.md
@@ -290,13 +290,21 @@ notifications:
 Backfill any missed intervals between start_date and now. Turn this on when you need to automatically recover historical
 runs after downtime or late onboarding.
 
+`catchup` accepts either a boolean or a string mode:
+
+- `false` (or omitted): no catchup
+- `true` or `"active"`: catch up only active assets
+- `"all"`: catch up every asset
+
+Any other string is treated as `false`. The value is always serialized as a string (`""`, `"active"`, or `"all"`).
+
 Example:
 
 ```yaml
-catchup: true
+catchup: active
 ```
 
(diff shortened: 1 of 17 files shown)


albertobruin:
Acknowledging the Greptile feedback — neither item changes the code:

1. **`null` JSON taking the bool branch.** As the comment notes, `json.Unmarshal("null", &b)` is a no-op leaving `b = false`, which yields `CatchupNone`. Since that's the same result the string branch would produce for `null`, this is just intent-clarity; I don't think the extra guard earns its keep.

2. **Silent normalization of unknown strings.** This is the explicit spec for the new field: any string outside `"active"`/`"all"` is treated as `false`. Erroring or warning on typos would re-introduce the validation behavior we're intentionally dropping. Happy to revisit if anyone disagrees.

batikankarakan (approved):

PR #2170, opened 2026-06-13
Support AWS default credential chain for secrets

### LINKED ISSUE #2152: AWS Secrets Manager backend should support the default credential chain (IAM roles), not only static BRUIN_AWS_* keys
### Summary

The AWS Secrets Manager secrets backend (`BRUIN_SECRETS_BACKEND=aws`) only authenticates with **static** credentials from `BRUIN_AWS_ACCESS_KEY_ID` / `BRUIN_AWS_SECRET_ACCESS_KEY` / `BRUIN_AWS_REGION`, and errors out if they aren't set. It can't use the **default AWS credential chain** (EC2 instance role, ECS task role, EKS IRSA, MWAA execution role, SSO, shared config), so on AWS infrastructure that already has an attached IAM role you're forced to inject static keys.

### Where (as of `37b82e5`)

`pkg/secrets/aws_secretsmanager.go` — `NewAWSSecretsManagerClientFromEnv` requires the env vars and hard-errors when missing:

https://github.com/bruin-data/bruin/blob/37b82e5fb09e7db955f7190b6b820ac3ed1faaf7/pkg/secrets/aws_secretsmanager.go#L33-L46

…and it builds the client with a **static** credentials provider:

https://github.com/bruin-data/bruin/blob/37b82e5fb09e7db955f7190b6b820ac3ed1faaf7/pkg/secrets/aws_secretsmanager.go#L65

By contrast, Bruin's *other* AWS integrations already use `config.LoadDefaultConfig(...)` (which honors the full credential chain, including instance/task roles):

- `pkg/s3/operator.go`
- `pkg/emr_serverless/operator.go`
- `pkg/quicksight/client.go`
- `pkg/config/connections.go`

So the capability exists in the codebase; the secrets backend just doesn't use it.

### Request

When `BRUIN_AWS_ACCESS_KEY_ID`/`BRUIN_AWS_SECRET_ACCESS_KEY` are **not** provided, fall back to `config.LoadDefaultConfig(ctx)` instead of erroring — consistent with the S3/EMR/QuickSight code paths. Suggested precedence:

1. Explicit `BRUIN_AWS_*` static creds (current behavior) — preserved as an override.
2. Otherwise, the default credential chain (env → shared config/SSO → container/instance role).

`BRUIN_AWS_REGION` could similarly fall back to the SDK's region resolution (`AWS_REGION` / shared config) when unset.

### Why it matters

Running Bruin on ECS / EKS / EC2 / MWAA with an attached IAM role is the standard production deployment. Today, to use Secrets Manager you must convert the ambient role's temporary credentials into `BRUIN_AWS_*` via an entrypoint shim (e.g. `aws configure export-credentials`), which is extra glue and re-implements what the SDK already does. Falling back to the default chain enables zero-config, role-based auth that also rotates automatically — and matches how the rest of Bruin's AWS features behave.

### Environment

Bruin CLI v0.11.618 (Windows), targeting an AWS ECS/MWAA deployment with an IAM task/execution role that already has `secretsmanager:GetSecretValue`.

karakanb:
Fixes #2152.

This lets the AWS Secrets Manager backend use the AWS SDK default credential and region chain when `BRUIN_AWS_*` static credentials are not provided. `BRUIN_AWS_ACCESS_KEY_ID` and `BRUIN_AWS_SECRET_ACCESS_KEY` still take precedence when both are set, and partial Bruin static credential config still errors.

Validated with `go test -tags="no_duckdb_arrow" ./pkg/secrets`, `make format`, `make test`, and `git diff --check`.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -1126,7 +1126,7 @@ func Run(isDebug *bool) *cli.Command {
 					errs = append(errs, errors.Wrap(err, "failed to initialize doppler client"))
 				}
 			case "aws":
-				connectionManager, err = secrets.NewAWSSecretsManagerClientFromEnv(logger)
+				connectionManager, err = secrets.NewAWSSecretsManagerClientFromEnv(ctx, logger)
 				if err != nil {
 					errs = append(errs, errors.Wrap(err, "failed to initialize AWS Secrets Manager client"))
 				}
--- a/pkg/secrets/aws_secretsmanager.go
+++ b/pkg/secrets/aws_secretsmanager.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/service/secretsmanager"
 	"github.com/bruin-data/bruin/pkg/config"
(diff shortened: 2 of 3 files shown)


mustafa3rsan (approved):

krom-als (approved):
May also be good to update the doc [https://github.com/bruin-data/bruin/blob/main/docs/secrets/aws-secrets-manager.md](https://github.com/bruin-data/bruin/blob/main/docs/secrets/aws-secrets-manager.md)

PR #2198, opened 2026-06-18
feat(cloud): add connections add/list/delete CLI commands

mustafa3rsan:


mustafa3rsan:
Closing in favor of a new approach: read connections from local .bruin.yml instead of requiring inline --credentials JSON.

PR #2208, opened 2026-06-19
feat(ingestr): add football-data.org as a source connection

iremcaginyurtturk:
## What

Adds **football-data.org** as an Ingestr source connection, so users can ingest FIFA World Cup soccer data (teams, stadiums, standings, matches, players, match events) into their warehouse via a `footballdata` connection.


iremcaginyurtturk:
Superseded by #2210, which is based directly on `main` (this one was stacked on the API-Football branch and mixed the two changes). Closing in favor of #2210.

PR #2263, opened 2026-06-26
Add SQL unit tests

krom-als:
**Size:** ~2,900 lines changed (`+2,851 / -66`) across 33 files — about 1,580 implementation, 1,040 tests, 290 docs.

## Files to review

Docs and tests aside, this is the implementation. The two files that carry the most weight are `cmd/unittest.go` (the command) and `pkg/unittest/warehouse.go` (how a test becomes one safe `SELECT`) — start there, the rest supports them.

**Command and wiring**
- `cmd/unittest.go` — the command: finds tests across pipelines, compiles and runs each one, derives the CTE-level checks
- `cmd/querier.go` — a small shared interface for "run this query, hand back typed rows"
- `cmd/fetch.go`, `cmd/import.go`, `cmd/patch.go` — moved onto that shared interface, no behavior change
- `main.go` — registers the command

**Asset schema and parsing**
- `pkg/pipeline/pipeline.go` — the `unit_tests` types on the asset model
- `pkg/pipeline/yaml.go` — reading `unit_tests:` out of the `@bruin` block

**Turning an asset into a test query**
- `pkg/unittest/warehouse.go` — swaps the tables an asset reads for fixture CTEs, freezes time, builds the final `SELECT`
- `pkg/unittest/fixtures.go` — resolves named/shared fixtures
- `pkg/unittest/compare.go` — compares rows to expectations, tolerant of how each warehouse returns numbers, dates, and column case

**SQL rewriting**
- `pkg/sqlparser/parser.go` — Go side of the new parser verbs (extract the inner `SELECT`, prepend CTEs, pull one CTE, freeze time)
- `pythonsrc/parser/main.py` — the sqlglot implementations, including the read-only gate that rejects writes
- `pythonsrc/parser/rename.py` — table-reference rewriting and the catalog handling that makes two- and three-part names work
- `pythonsrc/main.py` — dispatch for the new verbs

One config change to note: `.gitattributes` pins these files to LF endings.

## Tested against

I ran the full set of cases end to end against real warehouses, not just the Go unit tests:

- BigQuery
- Snowflake
- Postgres
- MySQL

Each run covered fixtures, CTE-level assertions, `execution_time` freezing, and three-part names where the platform supports them. The query builder is also checked against DuckDB in the Go test suite as an execution oracle.

## What this is

A new command, `bruin unit-test`, that checks the logic of a SQL asset against fixture data.

You hand a test some input rows for the tables an asset reads and the output you expect. Bruin rewrites the asset's query so those tables are swapped for your fixture rows, then runs the result as a single `SELECT` on the asset's own connection. Nothing is written. No temp tables, no inserts, no DDL.

The read-only guarantee isn't a connection setting (those don't behave the same across warehouses) — it's built into how the query is compiled. If an asset wraps its query in a `CREATE`, the test peels that off and runs the inner `SELECT`. If the SQL contains an actual write (`INSERT`/`UPDATE`/`DELETE`/`MERGE`), the test refuses to run it.

Because tests run on the real connection, you're checking against the exact SQL dialect that runs in production, not a local approximation of it.

## Running it

Same shape as `bruin validate`, so it drops straight into CI:

```
bruin unit-test                 # every pipeline under the current directory
bruin unit-test ./pipeline      # one pipeline
bruin unit-test ./assets/x.sql  # a single asset
```

Point it at the repo root and it finds every test on its own. No path argument required.

## Writing a test

Tests live in the asset definition. Here's one that checks refunds stay out of a revenue total:

```yaml
name: analytics.revenue_summary
type: sf.sql
connection: snowflake
unit_tests:
  - name: refunds_excluded
    inputs:
      - asset: analytics.orders
        rows:
          - {id: 1, status: paid,     amount: 100}
          - {id: 2, status: refunded, amount: 50}
    expected:
      rows:
        - {revenue: 100}
```

```sql
SELECT SUM(amount) AS revenue
FROM analytics.orders
WHERE status = 'paid'
```

You only assert the columns you care about; any extra columns in the result are ignored.

You can also assert on an intermediate CTE. That's useful when the final number is wrong and you want to know which step broke instead of guessing:

```yaml
unit_tests:
  - name: cte_paid_orders
    inputs:
      - asset: analytics.orders
        rows:
          - {id: 1, status: paid,     amount: 100}
          - {id: 2, status: paid,     amount: 50}
          - {id: 3, status: refunded, amount: 999}
    expected:
      ctes:
        paid_orders:
          rows:
            - {id: 1, amount: 100}
            - {id: 2, amount: 50}
      rows:
        - {revenue: 150}
```

```sql
WITH paid_orders AS (
  SELECT id, amount FROM analytics.orders WHERE status = 'paid'
)
SELECT SUM(amount) AS revenue FROM paid_orders
```

## Also supported

- Named fixtures defined once at the pipeline level and reused across tests
- Exact vs subset matching, row-count assertions, and strict row ordering
- Freezing `execution_time` so `CURRENT_DATE` and `CURRENT_TIMESTAMP` are deterministic
- Auto-stubbing for refs you didn't give rows for, so a test doesn't fall over on an unrelated upstream
- Two- and three-part table names

## Notes

- A command reference lives at `docs/commands/unit-test.md`, and the full authoring guide is under Data Governance at `docs/quality/unit-tests.md`.
- The fixture data is compiled into portable `SELECT ... UNION ALL` CTEs rather than `VALUES`, which is what makes the same test run unchanged on BigQuery through Postgres.

---


karakanb:
@greptileai

karakanb (approved):

PR #2272, opened 2026-06-30
feat(starrocks): add StarRocks as a Bruin source connection

iremcaginyurtturk:
## Summary

Adds **StarRocks** as a Bruin source connection, wiring the ingestr StarRocks source end-to-end. StarRocks federates open lakehouse table formats (Apache Iceberg, Hudi, Hive, Delta Lake) through external catalogs addressed as `catalog.database.table`, so those tables are readable through this connection too.

## Connection fields

`StarRocksConnection` exposes every parameter of the ingestr URI `starrocks://user:pass@host:port/[catalog/]database?ssl=...`:

| Field | Required | Notes |
|-------|----------|-------|
| `host` | ✅ | FE hostname/IP |
| `username` | ✅ | |
| `port` | optional | defaults to `9030` (FE MySQL protocol port) |
| `password` | optional | |
| `database` | optional | default database for unqualified tables |
| `catalog` | optional | external lakehouse catalog; prefixed only alongside a database, matching ingestr's `[catalog/]database` path rule |
| `ssl` | optional | `true` (verify) or `skip-verify` |

### DIFF
--- a/integration-tests/expectations/expected_connections_schema.json
+++ b/integration-tests/expectations/expected_connections_schema.json
@@ -1293,6 +1293,12 @@
           },
           "type": "array"
         },
+        "starrocks": {
+          "items": {
+            "$ref": "#/$defs/StarRocksConnection"
+          },
+          "type": "array"
+        },
         "dremio": {
           "items": {
             "$ref": "#/$defs/DremioConnection"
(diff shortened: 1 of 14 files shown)


integration-tests/expectations/expected_connections_schema.json
```diff
@@ -4052,6 +4058,41 @@
...
+          "type": "integer"
+        },
+        "database": {
+          "type": "string"
+        },
+        "catalog": {
+          "type": "string"
+        },
+        "ssl": {
+          "type": "string"
```
DjamilaBaroudi, line 4085: is this not boolean ?? if it's boolean vscode will display a choice in the dropdown with true and false values, but if it's string then it'll show an input

iremcaginyurtturk: this parameter can take different values: "true", "false" or "skip-verify". So it s defined as string

DjamilaBaroudi (approved):

PR #2385, opened 2026-07-15
starrocks: take partition and distribution from materialization

turtleDev:
StarRocks carried its own `distributed_by` and `partition_by` fields under the `starrocks` asset config block, duplicating notions that Bruin already models at the materialization level and that every other platform reads from there.

This drops both StarRocks-specific fields and sources the DDL from the shared materialization config instead:

- `PARTITION BY (...)` now comes from `materialization.partition_by` (a free-form string), so expression partitioning such as `date_trunc('day', ts)` works in addition to a plain column.
- `DISTRIBUTED BY HASH(...)` now comes from `materialization.cluster_by`, the same "columns that decide row layout" concept exposed on the other platforms, still falling back to the key columns when unset.

Only StarRocks-specific layout with no materialization equivalent (`table_model`, `buckets`, `properties`) remains under `starrocks`. Doris is left untouched.

## Example

```yaml
materialization:
  type: table
  strategy: create+replace
  cluster_by: [account_id]      # DISTRIBUTED BY HASH(account_id)
  partition_by: event_date      # PARTITION BY (event_date)

starrocks:
  table_model: primary_key
  buckets: 8
  properties:
    replication_num: "1"
```

### DIFF
--- a/docs/platforms/starrocks.md
+++ b/docs/platforms/starrocks.md
@@ -61,6 +61,8 @@ so they work the same way as on the other platforms:
 - `materialization.cluster_by` → `DISTRIBUTED BY HASH(...)` (defaults to the key columns)
 - `materialization.partition_by` → `PARTITION BY (...)` (a column or expression such as `date_trunc('day', event_date)`)
 
+`partition_by` is emitted verbatim so that expressions work, so it is **not** automatically backtick-quoted. If you partition on a single column whose name is a StarRocks reserved word (e.g. `date`, `value`, `key`), quote it yourself: `partition_by: "`date`"`.
+
 StarRocks-specific layout that has no materialization equivalent is declared under `starrocks`:
 
 ```yaml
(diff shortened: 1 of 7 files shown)


turtleDev:
Thanks for the review. Addressed in the latest commit:

- **Issue 1 — unquoted `partition_by`:** Intentional. The switch to a free-form string is what enables expression partitioning (`date_trunc('day', ts)`), so the value is emitted verbatim, consistent with how BigQuery and ClickHouse handle `partition_by`. Auto-quoting would break the expression case. Added a docs note that single reserved-word column names (`date`, `value`, `key`) must be backtick-quoted by the user.
- **Issue 3 — no expression test:** Added a DDL test case covering `partition_by: date_trunc('day', event_ts)`.
- **Issue 2 — migration shim for old `starrocks.distributed_by`/`partition_by`:** The non-strict YAML mechanism is real, but this config surface has never shipped in a release (the StarRocks platform commit is not contained in any tag), so there are no existing pipelines to migrate. Adding a permanent deprecation shim for a config that only ever existed on an unreleased branch would be dead weight — cleaning it up pre-release is exactly the right time. Declining this one.

batikankarakan (approved):

PR #2436, opened 2026-07-22
ci: use GitHub-native docker build for fork PRs

turtleDev:
## What

Split the `dockerBuild` job in the CI Checks workflow into two mutually exclusive jobs:

- **`dockerBuild`** (Depot, unchanged) now runs only for pushes to `main` and internally-created PRs.
- **`dockerBuildFork`** (new) runs only for PRs opened from forks, building the image with the GitHub-native `docker/setup-buildx-action` + `docker/build-push-action` on a stock `ubuntu-latest` runner.

## Why

Fork PRs cannot use Depot: the `depot/build-push-action` step needs an OIDC `id-token: write` token exchange against the Depot project, but workflows triggered from forks only receive a read-only `GITHUB_TOKEN` that cannot be elevated. This caused the Docker Build job to fail on every fork PR with a Depot authorization error.

The two jobs are gated on `github.event.pull_request.head.repo.full_name` vs `github.repository`, so exactly one runs per event. The fork job drops the registry login and `packages: write` permission since it only builds (`push: false`) and forks have no access to those secrets anyway.

## Note

If "Docker Build" is a required status check on `main`, add "Docker Build (fork)" to the required checks so fork PRs can satisfy branch protection.


DjamilaBaroudi (dismissed):

turtleDev:
Addressed the Greptile finding: the fork job now reports the same `Docker Build` check name as the Depot job (via the `name:` field). Since the two jobs are mutually exclusive and a skipped `if` job counts as passing for branch protection, a required `Docker Build` check is satisfied whichever path runs — no branch-protection change needed.

TanayBensuYurtturk (approved):

PR #2492, opened 2026-08-03
feat(iceberg): add ssl_mode to the postgres catalog

iremcaginyurtturk:
## Problem

Managed Postgres refuses plaintext connections, so an Iceberg **postgres catalog** on Neon, RDS or Cloud SQL needs `sslmode` set. Today the only way to say so is the free-form passthrough:

```yaml
catalog:
  type: postgres
  host: ep-….neon.tech
properties:
  sslmode: "require"
```

Two problems with that:

- **Every other connection has it as a field.** `postgres`, `redshift` and `mysql` all expose `ssl_mode`, so a user configuring an Iceberg postgres catalog reasonably looks for one, finds nothing, and gets a connection error that says nothing about TLS.
- **It cannot travel any other way.** The `sql` catalog can put `?sslmode=require` in its uri; the postgres catalog is configured by host and port, so the passthrough is the only route.

It also can't be surfaced in Bruin Cloud's connection form, since the form is generated from these fields — the user has to know to type a magic key into a properties box.

## Changes

`IcebergCatalog.SSLMode`, spelled **`ssl_mode`** to match the other connections, emitted as libpq's **`sslmode`**, which ingestr copies into the catalog connection string along with the other DSN parameters it recognises (`isPostgresDSNParam`).

```yaml
catalog:
  type: postgres
  host: ep-….neon.tech
  ssl_mode: require        # was: properties.sslmode
```

`properties` still wins on conflict, so existing configs are unaffected.

**Only the postgres catalog needs this.** The `sql` catalog carries TLS inside its uri, `rest` already has `rest_use_ssl`, `hive` speaks thrift, and sqlite/glue/hadoop have no connection to secure. The remaining libpq TLS parameters — `sslcert`, `sslkey`, `sslrootcert` — are file paths, which cannot exist on a hosted worker anyway, so they stay in `properties`.

### DIFF
--- a/docs/ingestion/iceberg.md
+++ b/docs/ingestion/iceberg.md
@@ -96,7 +96,7 @@ Each catalog type takes different fields. Use the matching `catalog:` block belo
 
 A managed database (Neon, RDS, Cloud SQL) refuses plaintext connections, so `ssl_mode` is usually required there. It takes libpq's values — `disable`, `allow`, `prefer`, `require`, `verify-ca`, `verify-full` — and is sent to the catalog database as `sslmode`.
 
-The remaining PostgreSQL connection parameters have no dedicated field; set them through the top-level [`properties`](#table-options) block, which ingestr forwards to the catalog database connection:
+`ssl_mode` is the only PostgreSQL connection parameter with a dedicated field. To set any of the others — certificate paths, timeouts, and so on — use the top-level [`properties`](#table-options) block, which ingestr forwards to the catalog database connection:
 
 ```yaml
           catalog:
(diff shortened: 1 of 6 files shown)


iremcaginyurtturk:
Closing — not worth a field. ingestr already forwards libpq parameters from the Iceberg connection's `properties`, so `ssl_mode` on the catalog would be a second spelling of `properties.sslmode` carried through config, URI builder and docs.

The real problem was discoverability, and that is a UI problem: bruin-data/cloud#2844 keeps the SSL Mode dropdown on the connection form and resolves it into `properties.sslmode` before it reaches the CLI, the same way it fills in `allow-unsafe-commits` for hadoop. Nothing new to add here, and it works against the current release.

PR #2676, opened 2026-09-09
fix(pipeline): enforce variable enum, min, and max constraints

### LINKED ISSUE #2587: Pipeline variable schemas (enum, minimum, maximum) are accepted but never enforced
## Summary

Pipeline variables accept JSON Schema keywords like `enum`, `minimum`, and `maximum`, and the docs show them being used — but nothing validates against them. A variable declared `enum: [reachable, contracted]` happily accepts `"Contracted"`, and the run succeeds. Values supplied with `--var` skip validation entirely.

## What happens

`Variables.Validate` (`pkg/pipeline/variables.go:19`) checks one thing: that each variable has a `default`. The schema compiler that would enforce everything else is commented out:

```go
// TODO(turtledev):
// - validate the defaults actually satisfy the schema
//
// BUG: Schema compiler fetches the schema from the spec URL, which may break
// in environments where internet access is restricted.

// TODO: Use a local copy of the schema.
// _, err := varSchemaLoader().Compile(gojsonschema.NewGoLoader(v.Schema()))
```

`Variables.Merge` (`pkg/pipeline/variables.go:72`), which backs `--var`, checks only that the variable name exists before overwriting the default:

```go
func (v *Variables) Merge(other map[string]any) error {
	for key, value := range other {
		if _, ok := (*v)[key]; !ok {
			return fmt.Errorf("no such variable %q", key)
		}
		(*v)[key]["default"] = value
	}
	return nil
}
```

So the constraint keywords are accepted, stored, and ignored.

## Reproducing

```yaml
variables:
  seat_denominator:
    type: string
    enum: [reachable, contracted]
    default: reachable
```

```bash
bruin run --var seat_denominator='"Contracted"' my-pipeline
```

```
 ✓ Assets executed      6 succeeded
 ✓ Quality checks      33 succeeded
```

`"Contracted"` is not in the enum. It is accepted, and the asset renders with a value that matches neither branch of the `{% if %}` that consumes it.

The same applies to numeric bounds — a variable declared `minimum: 7, maximum: 90` accepts `0` and `9999` without complaint.

## Why it matters

The failure is silent and the run is green, so the damage depends entirely on how the variable is consumed. In the case that prompted this, the variable selects between two ways of calculating the denominator of a score:

```sql
{% if var.seat_denominator == 'reachable' %}
  ...
{% elif var.seat_denominator == 'contracted' %}
  ...
{% endif %}
```

A user who deliberately switched to `contracted` and typed it with a capital letter got the setting they were trying to change away from, applied to every row, with no indication anything was wrong. The numbers stayed entirely plausible — just answering a different question than the one asked for.

Templates that declare bounds are getting no protection from them today. This includes `google-web-analytics`, which ships several (`source_lookback_days`, `reporting_window_days`, `trend_window_days`, `min_query_impressions`, `min_page_impressions`), and the docs in `docs/pipelines/definition.md` present `enum` and `minimum` as working features.

## Suggested fix

Validate values against the declared schema in two places:

1. **At parse time**, that each `default` satisfies its own schema — the first TODO in the existing comment.
2. **In `Merge`**, that an incoming `--var` value satisfies the schema before it replaces the default. This is the one that matters most in practice, since command-line overrides are exactly where typos happen.

On the commented-out compiler: the note says it fetches the meta-schema from a spec URL, which is why it was disabled. `gojsonschema` can be given a local copy of the draft-07 meta-schema through a `SchemaLoader`, which avoids the network dependency the BUG comment describes.

A smaller version of this would still be a big improvement: validating just `enum` and `minimum`/`maximum` for scalar types would cover most real declarations without needing the full compiler.

## Workaround

For any variable where a wrong value produces plausible-but-wrong output rather than an obvious break, fail inside the SQL:

```sql
{% if var.seat_denominator == 'reachable' %}
  ...
{% elif var.seat_denominator == 'contracted' %}
  ...
{% else %}
  CAST(ERROR('seat_denominator must be "reachable" or "contracted", got '
    || '{{ var.seat_denominator }}') AS INT64) AS licensed_seats
{% endif %}
```

Which works, but it is per-variable, per-platform boilerplate for something the schema already declares.

## Environment

- `bruin` built from `main` (dev)
- Reproduced with a BigQuery pipeline; nothing platform-specific

siddiqueirshad:
## What
Pipeline variables accept JSON Schema keywords like `enum`, `minimum`, and `maximum`, and the docs present them as working features — but until now only a missing `default` was rejected. A value such as `--var seat_denominator='"Contracted"'` was stored and the run succeeded, so Jinja branches silently took the wrong path.

This PR validates defaults, `--var` overrides, and variant values against `type`, `enum`, `const`, `minimum`, and `maximum` locally. It does not compile a full Draft-07 meta-schema (that approach was merged in #2599 and reverted in #2605), so in-progress templates such as zoomcamp's `type: TODO` keep parsing.

Closes #2587.

## Changes
- `pkg/pipeline/variables.go`: check defaults in `Validate` (used by `bruin validate`) and incoming values in `Merge` (used by `--var` and variant application). Overrides are applied only after every key is valid. Empty-bodied variables (`foo:`) no longer panic on merge.
- `pkg/pipeline/variant.go`: variant overrides go through the same constraint checks as `--var`.
- Tests for the reported enum typo, out-of-range integers, atomic merge, YAML pipeline.yml decoding, and `--var` mutator failures.
- Docs: custom variables, variants, pipeline definition, and the posthog-bigquery template now describe the enforced keywords.

## How to test
1. `go test -tags="no_duckdb_arrow" ./pkg/pipeline/ -run "TestVariables|TestVariantSet_Validate"`
2. `go test -tags="no_duckdb_arrow" ./cmd/ -run TestVariableOverridesMutator`
3. `make format` / `make test` if you have the full toolchain.
4. Manual: in a pipeline with `enum: [reachable, contracted]`, run `bruin run --var seat_denominator='"Contracted"'` and confirm it fails before assets execute. `reachable` / `contracted` still succeed. `forecast_horizon_days=0` fails when `minimum: 7`.

## Risk & rollback
- **Risk:** Medium. Previously accepted invalid variable values will now fail validation. Existing templates in this repo were checked; zoomcamp's placeholder `type: TODO` remains permitted because unknown types are not compiled as JSON Schema.
- **Rollback:** Revert the merge commit.

## Checklist
- [x] Unit tests added or updated (`make test`)
- [ ] Builds and lint pass (`make build-no-duckdb`, `make format`)
- [ ] Integration tests pass if affected (`make integration-test`)
- [x] No unrelated changes included
- [x] Tested locally

## Security
- [x] No secrets, credentials, or PII in this change
- [x] No new dependencies with known vulnerabilities
- [x] Security implications considered (if applicable)

## Related issues
Closes #2587

### DIFF
--- a/cmd/const_test.go
+++ b/cmd/const_test.go
@@ -214,4 +214,35 @@ func TestVariableOverridesMutator_VariantWinsOnOverlap(t *testing.T) {
 		require.Error(t, err)
 		assert.ErrorContains(t, err, `no such variable "nope"`)
 	})
+
+	t.Run("invalid enum --var is rejected", func(t *testing.T) {
+		t.Parallel()
+		p := &pipeline.Pipeline{
+			Variables: pipeline.Variables{
+				"seat_denominator": {"type": "string", "enum": []any{"reachable", "contracted"}, "default": "reachable"},
+			},
+		}
+
+		mutator := variableOverridesMutator([]string{`seat_denominator="Contracted"`})
+		_, err := mutator(t.Context(), p)
+		require.Error(t, err)
+		assert.ErrorContains(t, err, `invalid variable "seat_denominator"`)
+		assert.ErrorContains(t, err, "not one of the allowed values")
+		assert.Equal(t, "reachable", p.Variables["seat_denominator"]["default"])
+	})
+
+	t.Run("out of range --var is rejected", func(t *testing.T) {
+		t.Parallel()
+		p := &pipeline.Pipeline{
+			Variables: pipeline.Variables{
+				"forecast_days": {"type": "integer", "minimum": 7, "maximum": 90, "default": int64(30)},
+			},
+		}
+
+		mutator := variableOverridesMutator([]string{"forecast_days=0"})
+		_, err := mutator(t.Context(), p)
+		require.Error(t, err)
+		assert.ErrorContains(t, err, "below minimum")
+		assert.Equal(t, int64(30), p.Variables["forecast_days"]["default"])
+	})
 }
--- a/docs/getting-started/templates-docs/posthog-bigquery-README.md
+++ b/docs/getting-started/templates-docs/posthog-bigquery-README.md
@@ -515,12 +515,12 @@ the warehouse cannot speak to.
 
 ### Variable validation and macro loading
 
-The `enum`, `minimum`, and `maximum` keywords on these variables are currently
-documentation rather than enforcement — Bruin validates that a variable has a
-default and little else, and `--var` overrides skip schema checks. Where getting
+The `enum`, `minimum`, and `maximum` keywords on these variables are enforced
+on defaults, `--var` overrides, and variant values. A typo such as
+`--var seat_denominator='"Contracted"'` fails before any asset runs. Where getting
 a value wrong would silently produce plausible-but-wrong numbers rather than an
-obvious break, the SQL fails the run itself: an unrecognised `seat_denominator`
-stops `posthog_stage.accounts` with
+obvious break, the SQL still fails the run itself as a second line of defence: an
+unrecognised `seat_denominator` stops `posthog_stage.accounts` with
 `seat_denominator must be "reachable" or "contracted", got …` instead of quietly
 falling back to a default.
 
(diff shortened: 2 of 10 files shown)


siddiqueirshad:
Hi @turtleDev, when you get a chance, could you please take a look at this PR? Thanks!

